PR #6, opened 2018-12-13
Remove unused 'target' parameter from PromptForCredentials()

sclarke81:
It looks like `target` isn't used within `PromptForCredentials()` so it seems like a good idea to remove it unless it's there for a reason I've missed?


meziantou:
Thank you!

sclarke81:
No problem.

Issue #8, opened 2018-12-14
PromptForCredentials() returns null when the user enters a domain


sclarke81:
The logic in `GetCredentialsFromOutputBuffer()` doesn't handle domains correctly. I think there are two issues:

1. If `CredUnPackAuthenticationBuffer()` did return a domain then the outputs are set to null and false is returned.
1. If `CredUIParseUserName()` returns success the outputs are set to null and false is returned, even though a domain has been parsed.

sclarke81:
Do you think you could push a new NuGet package with this change in it?

meziantou:
The new package is available

sclarke81:
Awesome, thank you!

PR #9, opened 2018-12-14
#8 Change the logic in GetCredentialsFromOutputBuffer() so it handles domains

sclarke81:
Fix for #8


meziantou:
Thank you!

sclarke81:
No problem!

One thing that may or may not be of importance to you; I found that `CredUIParseUserName` only handles the `domain\username` and not the `username@domain` format.

PR #42, opened 2020-04-23
Add support for AddRange and RemoveRange

LePtitDev:


src/Meziantou.Framework.WPF/Collections/ConcurrentObservableCollection.cs
```diff
@@ -135,6 +149,20 @@ public bool Remove(T item)
             }
         }
 
+        public void RemoveRange(IEnumerable<T> items)
+        {
+            lock (_lock)
+            {
+                var itemsList = items.ToList();
```
meziantou: This doesn't need to be in the lock

tests/Meziantou.Framework.Tests/Collections/ConcurrentObservableCollectionTests.cs
```diff
@@ -0,0 +1,121 @@
+﻿using System.Windows.Threading;
```
meziantou: The tests should be in the project Meziantou.Framework.WPF.Tests

tests/Meziantou.Framework.Tests/Collections/ConcurrentObservableCollectionTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            // Act
+            using (collection.BeginBatch(BatchMode.Reset))
+            {
+                collection.AddRange(new[] { 1, 2, 3 });
+
+                // Assert
+                Assert.Equal(0, count);
+            }
+
+            Assert.Equal(1, count);
```
meziantou: It should validate the event is a reset event

tests/Meziantou.Framework.Tests/Collections/ConcurrentObservableCollectionTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            // Act
+            using (collection.BeginBatch())
+            {
+                collection.AddRange(new[] { 1, 2, 3 });
+
+                // Assert
+                Assert.Equal(0, count);
+            }
+
+            Assert.Equal(3, count);
```
meziantou: It should validate the events are add  events

src/Meziantou.Framework.WPF/Collections/BatchMode.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Meziantou.Framework.WPF.Collections
+{
+    public enum BatchMode
+    {
+        Default,
+        Optimized,
```
meziantou: I don't see any test for this mode

tests/Meziantou.Framework.WPF.Tests/Collections/ObservableCollectionTests.cs
```diff
@@ -143,6 +144,138 @@ public void AddWrongItemType()
...
+            // Arrange
+            var collection = new ConcurrentObservableCollection<int>(Dispatcher.CurrentDispatcher);
+
+            var events = new List<NotifyCollectionChangedEventArgs>();
+            collection.AsObservable.CollectionChanged += (sender, args) => events.Add(args);
+
+            // Act
+            using (collection.BeginBatch(BatchMode.Optimized))
+            {
+                collection.AddRange(new[] { 1, 2, 3 });
```
meziantou, line 244:
```
collection.AddRange(new[] { 1, 2, 3 });
collection.Remove(2);
collection.Add(2);
// Should combine the events, so the assert are still valid. But you should check the event Index so it match the actual collection indices.
```

src/Meziantou.Framework.WPF/Collections/Internals/DispatchedObservableCollection.cs
```diff
@@ -206,54 +232,112 @@ private void EnqueueEvent(PendingEvent<T> @event)
...
         {
+            foreach (var @event in events)
+            {
+                _pendingEvents.Enqueue(@event);
+            }
+
+            ProcessPendingEventsOrDispatch();
+        }
+
+        private void ProcessPendingEventsOrDispatch(bool forceProcess = false)
```
meziantou, line 245: I think it shoud be `processBatch`. BTW you should assert that `_batchMode` is not null in this case,

src/Meziantou.Framework.WPF/Collections/Internals/DispatchedObservableCollection.cs
```diff
@@ -354,5 +438,20 @@ void IList.RemoveAt(int index)
             AssertIsOnDispatcherThread();
             ((IList)_collection).RemoveAt(index);
         }
+
+        private class BatchModeScope : IDisposable
```
meziantou:
```suggestion
        private sealed class BatchModeScope : IDisposable
```

src/Meziantou.Framework.WPF/Collections/BatchMode.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Meziantou.Framework.WPF.Collections
+{
+    public enum BatchMode
+    {
+        Default,
```
meziantou: I think we should rename values
- DeferEvents
- CombineEvents
- ResetCollection

What do you think?

src/Meziantou.Framework.WPF/Collections/Internals/DispatchedObservableCollection.cs
```diff
@@ -206,54 +232,112 @@ private void EnqueueEvent(PendingEvent<T> @event)
...
+            {
+                RaiseResetEvent();
+            }
+        }
+
+        private IEnumerable<PendingEvent<T>> DequeueEvents()
+        {
+            if (_batchMode == BatchMode.Optimized)
+            {
+                var events = new List<PendingEvent<T>>();
```
meziantou: I'm not sure the Optimized flow works well with WPF. I think we should make more tests.

meziantou: I've looked at the code of the WPF controls. It seems the old value and index are important.
So, I think this is only possible to combine events that could be replaced by a Replace event or no event. For instance:

- Remove item, then insert a new item at this index => Replace the old value with the new one
- Replace the same item twice => One replace with the original value and the new value
- Add Last / Remove last => No event
- ...

Note: If you think this is too complex or don't want to do that, it's ok to only keep `BatchMode.Reset` in this PR. Then, you (or someone else) can create another PR to implement the optimized mode.

PR #63, opened 2020-09-20
Add .NET 5.0 target

meziantou:


src/Meziantou.Framework/ProcessExtensions.RunAsTask.cs
```diff
@@ -83,42 +111,47 @@ public static async Task<ProcessResult> RunAsTask(this ProcessStartInfo psi, Can
...
+                                catch (InvalidOperationException)
+                                {
+                                }
                             }
-                        }
-                    });
+                        });
+                    }
+
+                    await process.WaitForExitAsync(cancellationToken).ConfigureAwait(false);
```
meziantou, line 139: Should add WaitForExit?

PR #235, opened 2021-08-04
Add DisplayOrder to GenericForm

nibblersrevenge:
The [Display]-Annotation has the Property Order:

e.g.
 ```
       [Display(Name = "Lastname", Order = 1)]
        public string Name { get; set; }

        [Display(Name = "Firstname", Order = 2)]
        public string FirstName { get; set; }
```

### DIFF
--- a/src/Meziantou.AspNetCore.Components/GenericFormField.cs
+++ b/src/Meziantou.AspNetCore.Components/GenericFormField.cs
@@ -74,6 +74,7 @@ namespace Meziantou.AspNetCore.Components
                 return Property.Name;
             }
         }
+
         public int DisplayOrder
         {
             get
--- a/src/Meziantou.AspNetCore.Components/GenericFormField.cs
+++ b/src/Meziantou.AspNetCore.Components/GenericFormField.cs
@@ -74,6 +74,21 @@ namespace Meziantou.AspNetCore.Components
                 return Property.Name;
             }
         }
+        public int DisplayOrder
+        {
+            get
+            {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Meziantou.AspNetCore.Components/GenericFormField.cs [hunk 1]
```diff
@@ -75,6 +75,22 @@ public string DisplayName
...
+            {
+                var displayAttribute = Property.GetCustomAttribute<DisplayAttribute>();
+                if (displayAttribute != null)
+                {
+                    var displayOrder = displayAttribute.GetOrder();
+                    if (displayOrder != null)
+                        return (int)displayOrder;
+                }
+                //Undefined DisplayOrder are shown at the end 
+                return int.MaxValue;
```
meziantou: According to the documentation, we should use 10000 when no value is set

https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.getorder?view=net-5.0#remarks

```suggestion

                // https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.getorder?view=net-5.0#remarks
                // If an order is not specified, presentation layers should consider setting the value 
                // of the Order property to 10000. This value lets explicitly-ordered fields be displayed 
                // before and after the fields that do not have a specified order.
                return 10_000;
```

meziantou:
Thanks for your contribution 🚀🚀🚀 

I'll publish a new version of the package very soon

meziantou:
Package https://www.nuget.org/packages/Meziantou.AspNetCore.Components/1.0.10 is published

PR #236, opened 2021-08-06
Allow null values

nibblersrevenge:
GenericForm raises error while using e.g. int?

Changes allow  properties with typeof short?,int?,long?,float?,double?,decimal?

### DIFF
--- a/src/Meziantou.AspNetCore.Components/GenericFormField.cs
+++ b/src/Meziantou.AspNetCore.Components/GenericFormField.cs
@@ -74,6 +74,7 @@ namespace Meziantou.AspNetCore.Components
                 return Property.Name;
             }
         }
+
         public int DisplayOrder
         {
             get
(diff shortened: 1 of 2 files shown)


src/Meziantou.AspNetCore.Components/GenericFormField.cs
```diff
@@ -253,6 +253,24 @@ private static (Type ComponentType, IEnumerable<KeyValuePair<string, object>>? A
             if (property.PropertyType == typeof(decimal))
                 return (typeof(InputNumber<decimal>), null);
 
+            if (property.PropertyType == typeof(short?))
```
meziantou, line 256: I think it would be better to get the underlying type using `Nullable.GetUnderlyingType(property.PropertyType)` and use this value for the existing `if`. In the `return` statements, you can create the actual type using `typeof(InputNumber<>).MakeGenericType(property.PropertyType)`. This would prevent the duplication of `if`

meziantou: I've made the change in 4cd7a4a0e001ea6eaeb480cae953ea0bb2bea65b (I added you as a co-author of the commit 😊)

PR #237, opened 2021-08-06
Allow IEnumerable types in GenericForm

nibblersrevenge:
these changes allows IEnumerable as type.
It creates a &lt;SELECT&gt; in GenericForm and uses the KeyAttribute [KEY] from DataAnotations as identifier and the Property "Name" of &lt;TClass&gt;

### DIFF
--- a/Samples/BlazorAppSample/Pages/GenericFormDemo.razor
+++ b/Samples/BlazorAppSample/Pages/GenericFormDemo.razor
@@ -19,7 +19,7 @@
 <button type="button" @onclick="() => StateHasChanged()">Refresh</button>
 
 <div>
-@MyModel
+    @MyModel
 </div>
 
 @code {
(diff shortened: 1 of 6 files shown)


meziantou:
I don’t understand why the Customer class  would implement IEnumerable. It seems strange to me that a instance of customer can iterate on all instances. I would rather create a static method.

Also, you can already provide a custom editor that use a select (you can see the choice editor for example). I understand that you want a more automatic way of doing it. Maybe it would be a better design to create a custom editor that would provide values based on a static method (so you don’t have to implement an interface).

I think a better way would be to add an editor attribute on the Customer class. Then, the generic form could use it. This is not yet supported as the current code only checks the attribute declared on property. The improvement would be to also check if the attribute is declared on propertyinfo.propertytype. 

Using this approach, you would create a generic implementation of the editor in your code and declare it once on each class.

What do you think of this approach?

nibblersrevenge:
I think I understand what you mean. I created a version and a new pull request.

PR #238, opened 2021-08-08
InputCustomEditorSelect

nibblersrevenge:
Here is a new pull request, based on your comment of #237 
I hope that this version is better or more universal than the last. The naming is possibly still in need of improvement. 
What do you think?

### DIFF
--- a/Samples/BlazorAppSample/Pages/GenericFormDemo.razor
+++ b/Samples/BlazorAppSample/Pages/GenericFormDemo.razor
@@ -19,7 +19,7 @@
 <button type="button" @onclick="() => StateHasChanged()">Refresh</button>
 
 <div>
-@MyModel
+    @MyModel
 </div>
 
 @code {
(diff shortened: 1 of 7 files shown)


meziantou:
I don't think the `InputCustomEditorSelect` class should be part of the library. Indeed, it would require more design to support more scenario such as loading item asynchronouly. So, if you want to keep it in this code base, you can move it to the sample project.

My point from the other PR was to provide a simpler way to set the editor for a type. For instance, if you want to set a custom editor for all properties of type `Customer`. This is now possible (https://github.com/meziantou/Meziantou.Framework/commit/34d1cee40a6d387d0b992612512d166ce778c75a) by adding the `Editor` attribute on the `Customer` class:

````c#
[System.ComponentModel.Editor(typeof(InputCustomEditorSelect<Customer>), typeof(InputBase<>))]
public class Customer
{
    // ...
}
````

nibblersrevenge:
I understand why you made that decision and in principle I also agree with you.  👌

PR #493, opened 2022-11-10
Implemented 'IStronglyTypedId' interface and extension methods to identity 'StronglyTypedId' properties

jeffward01:
Hello!

For a long time I have admired your work and way you have built your framework to make building code 'easier' (less boilerplate, useful classes / helpers, etc).

I noticed in your project [StronglyTypedId](https://github.com/meziantou/Meziantou.Framework/tree/main/src/Meziantou.Framework.StronglyTypedId), that there was not an 'automatic way' to detect if a class property was indeed a `StronglyTypedId` or something else.

[Specifically looking here at your blog post](https://www.meziantou.net/strongly-typed-ids-with-csharp-source-generators.htm#entity-framework-cor) - When using `StronglyTypedId`s with Entity framework - each Id must be referenced manually within the `OnModelCreating` method.

I wanted to add a way to 'automatically detect' which classes where `StronglyTypedId` and which were not.

I saw that the StronglyTypedId had a base-type of `System.ValueType` - but I thought this could be used as an 'identifier' to identify `StronglyTypedId` vs some other class  - but I thought that `System.ValueType` was to generic.  

Instead I opted to use a 'marker interface' similar to what is used in Mediatr, which I named `IStronglyTypedId<>` - I realize that some people are against 'marker interfaces' and view them as 'junk code' - however in this case I thought it was acceptable.

---

## Purpose of Feature

The purpose behind this is to minimize the amount of 'boiler-plate code the user would need to manually implement themselves.

Many of my projects are larger enterprise projects with 100+ `StronglyTypedId` objects - I would prefer to not implement those by hand.

### Tasks

- [x] Create Generic Interface that allows `StronglyTypedId` (properties / types) to be identified easily
- [x] Unit Tests for `IStronglyTypedId<>`
- [x] Create Extension methods that allow the user to perform common identification method of 'StronglyTypedId' properties and their underlying 'value type'
- [x] Unit Tests for the extension methods
- [ ] Create generic type 'converters' with functionality for types [defined here](https://github.com/meziantou/Meziantou.Framework/blob/main/src/Meziantou.Framework.StronglyTypedId/StronglyTypedIdSourceGenerator.cs#:~:text=TypeReference%20GetTypeReference) (These will just be generic converters that are not specific to Entity Framework or MongoDb) (discussion would be nice)
- [ ] Create new project (nuget package) for Value Converters specific for Entity Framework?? (discussion would be nice)
- [ ] Create new project (nuget package) for Value Converters specific for MongoDb and others?? (discussion would be nice)

### Interface

Note: This is automatically generated by the source-generator 

````csharp

 public interface IStronglyTypedId<out TType>
     
{
    TType Value { get; }
}

````

## Extension Methods

Note: This is automatically generated by the source-generator 

````csharp


////////////////////////////////////////////////////////////////////////////////////////////////////
/// <summary>
///     Extension Methods to identify strongly-typed id's and extract the type of value object which is implemented
///     such as 'string', 'Guid', 'int', or other valid types.
/// </summary>
////////////////////////////////////////////////////////////////////////////////////////////////////
public static class StronglyTypedIdExtensions
{
    ////////////////////////////////////////////////////////////////////////////////////////////////////
    /// <summary>
    ///     Returns the type of value object which is implemented in the strongly-typed id value object.
    ///     <para>Such as 'string', 'Guid', 'int' or any other allowed value types. </para>
    /// </summary>
    /// <exception cref=""InvalidDataException"">
    ///     Throws when <paramref name=""stronglyTypedId"" /> is not a strongly-typed id value object
    /// </exception>
    /// <param name=""stronglyTypedId"">  The strongly-typed id to extract the base 'value type' from. </param>
    /// <returns>   Returns the 'Type' of value object</returns>
    ////////////////////////////////////////////////////////////////////////////////////////////////////
    public static Type GetTypeOfStronglyTypedId(this Type stronglyTypedId)
    {
        if (!stronglyTypedId.IsStronglyTypedId())
        {
            throw new InvalidDataException($""The type {stronglyTypedId.Name} is not a strongly typed id"");
        }

        return stronglyTypedId.GetTypeInfo()
            .GetFields(BindingFlags.NonPublic | BindingFlags.Instance)[0]
            .FieldType;
    }

    ////////////////////////////////////////////////////////////////////////////////////////////////////
    /// <summary>
    ///     Returns the type of value object which is implemented in the strongly-typed id value
    ///     object.
    ///     <para>Such as 'string', 'Guid', 'int' or any other allowed value types. </para>
    /// </summary>
    /// <remarks>   Jeff Ward, 11/10/2022. </remarks>
    /// <exception cref=""InvalidDataException"">
    ///     Thrown when an Invalid Data error condition occurs.
    /// </exception>
    /// <param name=""stronglyTypedPropertyInfo"">    The strongly-typed id property to extract the base 'value type' from. </param>
    /// <returns>   Returns the 'Type' of value object. </returns>
    ////////////////////////////////////////////////////////////////////////////////////////////////////
    public static Type GetTypeOfStronglyTypedId(this PropertyInfo stronglyTypedPropertyInfo)
    {
        if (!stronglyTypedPropertyInfo.IsStronglyTypedId())
        {
            throw new InvalidDataException($""The type {stronglyTypedPropertyInfo.Name} is not a strongly typed id property"");
        }

        return stronglyTypedPropertyInfo.PropertyType.GetTypeInfo()
            .GetFields(BindingFlags.NonPublic | BindingFlags.Instance)[0]
            .FieldType;
    }


    ////////////////////////////////////////////////////////////////////////////////////////////////////
    /// <summary>
    ///     Determines if this <paramref name=""type"" /> has any properties that are strongly-typed id value objects.
    /// </summary>
    /// <param name=""type""> Determines if this type contains any properties that are strongly-typed id value objects. </param>
    /// <returns>   Returns true if this type has properties that are strongly-typed id value objects. </returns>
    ////////////////////////////////////////////////////////////////////////////////////////////////////
    public static bool HasStronglyTypedIdProperty(this Type type)
    {
        return type.GetProperties(BindingFlags.Public | BindingFlags.Instance)
            .Any(_ => _.IsStronglyTypedId());
    }

    ////////////////////////////////////////////////////////////////////////////////////////////////////
    /// <summary>
    ///     Determines if the <seealso cref=""PropertyInfo"" /> <paramref name=""pi"" /> is a strongly-typed id value object
    /// </summary>
    /// <param name=""pi""> Determines if this property is a strongly-typed id value object. </param>
    /// <returns>  Returns true if the property is a strongly-typed id value object</returns>
    ////////////////////////////////////////////////////////////////////////////////////////////////////
    public static bool IsStronglyTypedId(this PropertyInfo pi)
    {
        return pi.PropertyType.IsStronglyTypedId();
    }

    ////////////////////////////////////////////////////////////////////////////////////////////////////
    /// <summary>
    ///     Determines if the <paramref name=""type"" /> is a strongly-typed id value object.
    /// </summary>
    /// <param name=""type""> Determines if this is a strongly-typed id value object. </param>
    /// <returns>  Returns true if the <paramref name=""type"" /> is a strongly-typed id value object.</returns>
    ////////////////////////////////////////////////////////////////////////////////////////////////////
    public static bool IsStronglyTypedId(this Type type)
    {
        return type.GetInterfaces()
            .Any(_ => _.Name == ""IStronglyTypedId"");
    }
}
````

------

### Entity Framework Example (Currently Implemented)

 
Now, users can do something like this:

````csharp
    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        foreach (var entityType in modelBuilder.Model.GetEntityTypes())
        {

            foreach(var prop in entityType.ClrType.GetProperties(BindingFlags.Public | BindingFlags.Instance))
            {
                if (prop.IsStronglyTypedId())
                {
                    var valueType = prop.GetTypeOfStronglyTypedId();

                    if (valueType is int intValueType)
                    {
                        // some conversion methods
                    }

                    // The can also do this
                    var propertyType = prop.PropertyType; //type -> ProjectId (strongly-typed id)

                    if (propertyType is ProjectId projectId)
                    {
                        // some conversion methods
                    }
                }

            }
        }

        base.OnModelCreating(modelBuilder);

````

[Here are the Microsoft docs for 'ValueConverter' within Entity Framework](https://learn.microsoft.com/en-us/ef/core/modeling/value-conversions?tabs=data-annotations)

Example Implementation from Microsoft:

````csharp

public class CurrencyConverter : ValueConverter<Currency, decimal>
{
    public CurrencyConverter()
        : base(
            v => v.Amount,
            v => new Currency(v))
    {
    }
}

````

Example StronglyTypedConverter
````csharp

[StronglyTypedId(typeof(Guid))]
public readonly partial struct ProductId  { }

//----------------------------------------------------------


  public class StronglyTypedGuidValueConverter : ValueConverter<IStronglyTypedId<Guid>, Guid>
    {
        public StronglyTypedGuidValueConverter()
            : base(
                v => v.Value,
                v => ProductId.FromGuid(v))
        {
        }
    }

````

I don't love this current solution because it still requires some form of manual implementation.  What I could do is some form of 'interface implementation' based on these allowed types you have below.  Basically it will map generically based on the interface of type: `IStronglyTypedId<>`

````csharp
    private static TypeReference GetTypeReference(IdType type)
    {
        return type switch
        {
            IdType.System_Boolean => new TypeReference(typeof(bool)),
            IdType.System_Byte => new TypeReference(typeof(byte)),
            IdType.System_DateTime => new TypeReference(typeof(DateTime)),
            IdType.System_DateTimeOffset => new TypeReference(typeof(DateTimeOffset)),
            IdType.System_Decimal => new TypeReference(typeof(decimal)),
            IdType.System_Double => new TypeReference(typeof(double)),
            IdType.System_Guid => new TypeReference(typeof(Guid)),
            IdType.System_Int16 => new TypeReference(typeof(short)),
            IdType.System_Int32 => new TypeReference(typeof(int)),
            IdType.System_Int64 => new TypeReference(typeof(long)),
            IdType.System_SByte => new TypeReference(typeof(sbyte)),
            IdType.System_Single => new TypeReference(typeof(float)),
            IdType.System_String => new TypeReference(typeof(string)),
            IdType.System_UInt16 => new TypeReference(typeof(ushort)),
            IdType.System_UInt32 => new TypeReference(typeof(uint)),
            IdType.System_UInt64 => new TypeReference(typeof(ulong)),
            _ => throw new ArgumentException("Type not supported", nameof(type)),
        };
    }
````

### CODE REVIEW

**Comment by jeffward01** [comment]
I wanted to get your opinion specifically  on:

- [ ] Create generic type 'converters' with functionality for types [defined here](https://github.com/meziantou/Meziantou.Framework/blob/main/src/Meziantou.Framework.StronglyTypedId/StronglyTypedIdSourceGenerator.cs#:~:text=TypeReference%20GetTypeReference) (These will just be generic converters that are not specific to Entity Framework or MongoDb) (discussion would be nice)


### Goals:

If a user has 150 StronglyTypedId's as an example, they should not need to implement 150 converters, or manually write 150 lines of code like:

````csharp
modelBuilder.Entity(entityType.Name).Property(prop.Name)
.HasConversion(new ValueConverter<CustomerId, int>(c => c.Value, c => CustomerId.FromInt32(c)));

`````

Conversion should be 'near seamless' and easy - no manual code conversion would be the goal

-----

I plan to do some sort of interface implementation which will allow conversion to and from StronglyTypedId types generically based on:

````csharp
 public interface IStronglyTypedId<out TType> 
{
    TType Value { get; }

    IStronglyTypedId<TType> ConvertFrom(Type valueType);

    TType ConvertTo(IStronglyTyped<Type> stronglyTypedId)
}
````

or perhaps a base class with protected methods.

Then I would implement some converters that would have functionality to convert these types [defined here](https://github.com/meziantou/Meziantou.Framework/blob/main/src/Meziantou.Framework.StronglyTypedId/StronglyTypedIdSourceGenerator.cs#:~:text=TypeReference%20GetTypeReference)

-----

## Questions

I wanted to ask your opinion:

1.) On my pull-request (not complete) 
2.) The 'automatic type converters' (above)

I might be over-complicating things, and perhaps the solution is simple.  I just would love your opinion on it.

Thanks
Jeff

**Comment by jeffward01** [comment]
I think this: [Using strongly-typed entity IDs with EF Core - Part 3](https://andrewlock.net/using-strongly-typed-entity-ids-to-avoid-primitive-obsession-part-3/) and his other articles on this topic is a good solution for the 'type conversation' topic.

He has parts 3-7 which seem to sound like a good solution.

If you agree that this is the best route I can get started on implementing a similar functionality.

If you have any objections i'd love to hear them to learn more

Thanks!

**Comment by meziantou** [comment]
> I noticed in your project [StronglyTypedId](https://github.com/meziantou/Meziantou.Framework/tree/main/src/Meziantou.Framework.StronglyTypedId), that there was not an 'automatic way' to detect if a class property was indeed a StronglyTypedId or something else.

The type is decorated with an attribute, so you already have a way to know if a type is a strongly-typed id. Why do you need an additional way to detect these types?

> I saw that the StronglyTypedId had a base-type of System.ValueType - but I thought this could be used as an 'identifier' to identify StronglyTypedId vs some other class - but I thought that System.ValueType was to generic.

If you use the attribute on a struct / struct record, then this is true. But you can use the attribute on a class or a record.

> Instead I opted to use a 'marker interface' similar to what is used in Mediatr, which I named IStronglyTypedId<> - I realize that some people are against 'marker interfaces' and view them as 'junk code' - however in this case I thought it was acceptable.

I'm not against the interface if it adds value. However, this interface cannot be public if it is generated by the source generator. Otherwise, each assembly will have its own version of the interface and there will be conflicts.

> Create Extension methods that allow the user to perform common identification method of 'StronglyTypedId' properties and their underlying 'value type'

I don't see a reason to generate this code. These methods can be implemented in another assembly.

> I plan to do some sort of interface implementation which will allow conversion to and from StronglyTypedId types generically based on:

Question in the code:

````
public interface IStronglyTypedId<out TType> 
{
    // Why do you prefer a property instead of using the type from the StronglyTypedAttribute?
    TType Value { get; }

    // Looks very similar to the already implemented feature: System.ComponentModel.TypeConverter
    IStronglyTypedId<TType> ConvertFrom(Type valueType);
    TType ConvertTo(IStronglyTyped<Type> stronglyTypedId)
}
````

> EF Core

For EF Core, I think the generator could generate an extension method like

````
internal static void MapStonglyTypedIds(this ModelBuilder modelBuilder);
````

However, how do you handle Ids from multiple assemblies?

**Comment by rafalschmidt97** [comment]
Haven't read the whole thread, but for some reason we've made our own abstraction on top of this library:

```
https://github.com/allegro/dotnet-utils/blob/main/src/Allegro.Extensions.Identifiers/Allegro.Extensions.Identifiers.Abstractions/IStronglyTypedId.cs
```

meziantou:
I've added the support for interfaces in https://github.com/meziantou/Meziantou.Framework/pull/519. If you add a reference to the `Meziantou.Framework.StronglyTypedId.Interfaces` package, the source generator will add the interfaces to the generated types.

PR #512, opened 2023-02-16
fix boolean mistake in ToString of nullable idType

pschwarzpp:
`someIdType.ToString()` returns `SomeIdType { Value = <null> }` when actually a value is not null

### DIFF
--- a/src/Meziantou.Framework.StronglyTypedId/StronglyTypedIdSourceGenerator.Members.cs
+++ b/src/Meziantou.Framework.StronglyTypedId/StronglyTypedIdSourceGenerator.Members.cs
@@ -76,11 +76,11 @@ public partial class StronglyTypedIdSourceGenerator
                 {
                     using (writer.BeginBlock($"if ({PropertyName} == null)"))
                     {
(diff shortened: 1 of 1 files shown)


src/Meziantou.Framework.StronglyTypedId/StronglyTypedIdSourceGenerator.Members.cs
```diff
@@ -76,11 +76,11 @@ string ValueToStringExpression()
                 {
                     using (writer.BeginBlock($"if ({PropertyName} == null)"))
                     {
-                        writer.WriteLine($$"""return "{{context.TypeName}} { Value = " + {{PropertyAsStringName}} + " }";""");
+                        writer.WriteLine($$"""return "{{context.TypeName}} { Value = <null> }";""");
```
meziantou, line 79: Could you add a test in https://github.com/meziantou/Meziantou.Framework/blob/main/tests/Meziantou.Framework.StronglyTypedId.GeneratorTests/StronglyTypedIdTests.cs#L14 to validate the result of `ToString`?

(note: you may need to run the test on this project using the command line `dotnet test` as VS is not able to correctly load the source generator built from the referenced project)

PR #513, opened 2023-02-18
Fix loading state

evilz:
_loading field was not set before and after loading.

### DIFF
--- a/src/Meziantou.AspNetCore.Components/InfiniteScrolling/InfiniteScrolling.razor
+++ b/src/Meziantou.AspNetCore.Components/InfiniteScrolling/InfiniteScrolling.razor
@@ -66,6 +66,7 @@
         _loadItemsCts = cts;
         try
         {
+            _loading = true;
             StateHasChanged(); // Allow the UI to display the loading indicator
             try
             {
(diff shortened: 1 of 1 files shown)


src/Meziantou.AspNetCore.Components/InfiniteScrolling/InfiniteScrolling.razor
```diff
@@ -66,6 +66,7 @@
         _loadItemsCts = cts;
         try
         {
+            _loading = true;
```
meziantou, line 69: In fact, the `_loading` field should not exist. It should have been replaced with `private bool IsLoading => _loadItemsCts != null;`. Do you want to do the change?

note: the bug was introduced in 5ca8395b87e71b2e6813f1ddb3ec69fc389fa994

meziantou: Thanks for reporting the bug. It is fixed by 7a3278545db69d6a44062d67628610cb98a688a9.

evilz: > In fact, the `_loading` field should not exist. It should have been replaced with `private bool IsLoading => _loadItemsCts != null;`. Do you want to do the change?
> 
> note: the bug was introduced in 5ca8395b87e71b2e6813f1ddb3ec69fc389fa994

This also what I had in mind, but was not sure you expect to do it like this. Perfect.

PR #526, opened 2023-04-12
CodeOwners inline comment fix

jamesc-skyward:
When I parsed GitHub's example (https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file) it didn't ignore the inline comment. It instead treated each word in the comment as a member.

This example
```
# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
*.js    @js-owner #This is an inline comment.
```
Resulted in 6 owners ("js-owner", "#This", "is", "an", "inline", "comment.").

With this change, it should now kick out of `ParseMembers` when it comes across a # character.


meziantou:
Thanks!

PR #541, opened 2023-06-27
Retry HTTP request with delay hint (on 429)

### LINKED ISSUE #539: Consider retries in `SymbolsValidationRule` due to intermittent HTTP errors
[I'm using](https://github.com/morelinq/MoreLINQ/blob/a9437d48e3d30cf05035853d2e22226f630a2468/appveyor.yml#L89) your excellent [package validation tool](https://www.nuget.org/packages/Meziantou.Framework.NuGetPackageValidation.Tool) (thanks!) in the CI builds for [MoreLINQ](https://morelinq.github.io/). Recently, I've noticed the CI builds failing during package validation with the error code 119:

https://github.com/meziantou/Meziantou.Framework/blob/f35b1c30af34cab4cc75b08218d4b2e36d08533f/src/Meziantou.Framework.NuGetPackageValidation/Rules/ErrorCodes.cs#L53

An example of the output from the tool is:

```json
{
  "Packages": {
    "/home/appveyor/projects/morelinq/dist/morelinq.4.0.0-ci-20230627T0635.nupkg": {
      "Errors": [
        {
          "ErrorCode": 119,
          "Message": "Source file \u0027https://raw.githubusercontent.com/morelinq/MoreLINQ/a9437d48e3d30cf05035853d2e22226f630a2468/MoreLinq/ZipImpl.cs\u0027 is not accessible",
          "FileName": "lib/net6.0/MoreLinq.dll"
        }
      ]
    }
  }
}
```

The error seems to be intermittent and fails on different URLs at different images of the build each time. When I try to access the identified URL, it works. When I re-run the CI build, it usually passes. Would you consider adding a retry? If so, I'd be glad to send a PR with an implementation.

atifaziz:
@meziantou [said](https://github.com/meziantou/Meziantou.Framework/issues/539#issuecomment-1609928285):

>  I think it could be improved by handling correctly the status code 429 (I would accept a PR).

This PR contributes to and potentially closes #539.

The value of the [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is used as long as it's present and we're retrying. In the even that there's no retry duration hint, it uses the default like before. If you think the default could revised to be less aggressive then I can make that change in a follow-up commit. My suggestion would be starting with a second and doubling on each attempt.

Let me know if anything needs to be adapted to the project's coding style although I tried to respect what I could see from going through a few of sources.

### DIFF
--- a/src/Meziantou.Framework.NuGetPackageValidation/Internal/SharedHttpClient.cs
+++ b/src/Meziantou.Framework.NuGetPackageValidation/Internal/SharedHttpClient.cs
@@ -24,9 +24,10 @@ internal static class SharedHttpClient
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
             const int MaxRetries = 5;
-            for (var i = 1; ; i++)
+            var defaultDelay = TimeSpan.FromMilliseconds(200);
+            for (var i = 1; ; i++, defaultDelay *= 2)
             {
-                TimeSpan? delay = null;
+                TimeSpan? delayHint = null;
                 HttpResponseMessage? result = null;
 
                 try
@@ -38,7 +39,7 @@ internal static class SharedHttpClient
                         // either a 503 (Service Unavailable) or 429 (Too Many Requests):
                         // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After
 
-                        delay = result.Headers.RetryAfter switch
+                        delayHint = result.Headers.RetryAfter switch
                         {
                             { Date : { } date  } => date - DateTimeOffset.UtcNow,
                             { Delta: { } delta } => delta,
@@ -59,7 +60,7 @@ internal static class SharedHttpClient
                         throw;
                 }
 
-                await Task.Delay(delay is { } someDelay && someDelay > TimeSpan.Zero ? someDelay : TimeSpan.FromMicroseconds(200 * i), cancellationToken).ConfigureAwait(false);
+                await Task.Delay(delayHint is { } someDelay && someDelay > TimeSpan.Zero ? someDelay : defaultDelay, cancellationToken).ConfigureAwait(false);
 
                 static bool IsLastAttempt(int i) => i >= MaxRetries;
             }
--- a/src/Meziantou.Framework.NuGetPackageValidation/Internal/SharedHttpClient.cs
+++ b/src/Meziantou.Framework.NuGetPackageValidation/Internal/SharedHttpClient.cs
@@ -26,13 +26,26 @@ internal static class SharedHttpClient
             const int MaxRetries = 5;
             for (var i = 1; ; i++)
             {
+                TimeSpan? delay = null;
                 HttpResponseMessage? result = null;
+
                 try
                 {
                     result = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
-                    if (!IsLastAttempt(i) && ((int)result.StatusCode >= 500 || result.StatusCode is System.Net.HttpStatusCode.RequestTimeout))
+                    if (!IsLastAttempt(i) && ((int)result.StatusCode >= 500 || result.StatusCode is System.Net.HttpStatusCode.RequestTimeout or System.Net.HttpStatusCode.TooManyRequests))
                     {
                         result.Dispose();
+
+                        // Use "Retry-After" value, if available. Typically, this is sent with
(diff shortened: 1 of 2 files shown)


src/Meziantou.Framework.NuGetPackageValidation/Internal/SharedHttpClient.cs
```diff
@@ -26,13 +26,26 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
...
-                    if (!IsLastAttempt(i) && ((int)result.StatusCode >= 500 || result.StatusCode is System.Net.HttpStatusCode.RequestTimeout))
+                    if (!IsLastAttempt(i) && ((int)result.StatusCode >= 500 || result.StatusCode is System.Net.HttpStatusCode.RequestTimeout or System.Net.HttpStatusCode.TooManyRequests))
                     {
                         result.Dispose();
+
+                        // Use "Retry-After" value, if available. Typically, this is sent with
+                        // either a 503 (Service Unavailable) or 429 (Too Many Requests):
+                        // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After
+
+                        delay = result.Headers.RetryAfter switch
```
meziantou: I don't know if this is safe to access `result` right after disposing it. I think you should move the `Dispose` call after this one.

src/Meziantou.Framework.NuGetPackageValidation/Internal/SharedHttpClient.cs
```diff
@@ -46,7 +59,7 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                         throw;
                 }
 
-                await Task.Delay(100 * i, cancellationToken).ConfigureAwait(false);
+                await Task.Delay(delay is { } someDelay && someDelay > TimeSpan.Zero ? someDelay : TimeSpan.FromMicroseconds(100 * i), cancellationToken).ConfigureAwait(false);
```
meziantou: I think you are right, the default delay is too aggressive. I think we can start with `TimeSpan.FromMilliseconds(200 * i)`. You can implement something smarter if you prefer.

meziantou (changes requested):

src/Meziantou.Framework.NuGetPackageValidation/Internal/SharedHttpClient.cs
atifaziz: > I don't know if this is safe to access `result` right after disposing it.

The [`Dispose` implementation of `HttpResponseMessage`](https://github.com/dotnet/runtime/blob/v7.0.8/src/libraries/System.Net.Http/src/System/Net/Http/HttpResponseMessage.cs#L212-L221) only [disposes the content](https://github.com/dotnet/runtime/blob/v7.0.8/src/libraries/System.Net.Http/src/System/Net/Http/HttpResponseMessage.cs#L219). It doesn't invalidate/reset the entire object's state. The [`Headers` getter makes no check about the disposed state of the object](https://github.com/dotnet/runtime/blob/v7.0.8/src/libraries/System.Net.Http/src/System/Net/Http/HttpResponseMessage.cs#L107). The [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpresponsemessage.dispose) doesn't call out anything special either. That all said, I agree that it may be best not to rely on those implementation details and so I've moved the `Dispose` call after with 692eacbc4929a835efab87ca12497967e2ebe17d.

src/Meziantou.Framework.NuGetPackageValidation/Internal/SharedHttpClient.cs
atifaziz: > I think we can start with `TimeSpan.FromMilliseconds(200 * i)`.

Done with e01f7d45af4cfd8b2e3fe683e3fbb13b572bdbc8.

atifaziz: > You can implement something smarter if you prefer.

I implemented a more exponential strategy with 692eacbc4929a835efab87ca12497967e2ebe17d. I can revert if you object. The [GitHub documentation recommends something similar](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#secondary-rate-limits):

> Otherwise, wait for an exponentially increasing amount of time between retries, and throw an error after a specific number of retries.

meziantou:
Thanks for your contribution! The new packages are available on NuGet

PR #553, opened 2023-08-24
Added XML comments on class and method in FastEnumToStringGenerator

petrkoutnycz:
I came across a situation where I need the generated `ToStringFast()` methods to be public but we have enabled and required XML comments on public members so the project is unable to build. So I am adding a generation for public classes and methods.

Unfortunately, I did not find a proper way on how to unit test generated XML comments so I did not extend existing tests.

### DIFF
--- a/src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
+++ b/src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
@@ -135,7 +135,9 @@ internal sealed class FastEnumToStringAttribute : System.Attribute
 
         foreach (var enumerationGroup in enums.GroupBy(en => en.FullNamespace, StringComparer.Ordinal).OrderBy(g => g.Key, StringComparer.Ordinal))
         {
-            var typeVisibility = enumerationGroup.Any(enumeration => enumeration.IsPublic) ? "public" : "internal";
+            var typeIsPublic = enumerationGroup.Any(enumeration => enumeration.IsPublic);
+            var typeVisibility = typeIsPublic ? "public" : "internal";
+
             foreach (var enumeration in enumerationGroup.OrderBy(e => e.FullCsharpName, StringComparer.Ordinal))
             {
                 var methodVisibility = enumeration.IsPublic ? "public" : "internal";
@@ -146,8 +148,18 @@ internal sealed class FastEnumToStringAttribute : System.Attribute
                     sb.AppendLine("{");
                 }
 
+                if (typeIsPublic)
+                {
+                    sb.AppendLine($"/// <summary>A class with memory-optimized alternative to regular ToString() on enums.</summary>");
+                }
+
                 sb.Append(typeVisibility).AppendLine(" static partial class FastEnumToStringExtensions");
                 sb.AppendLine("{");
+
+                if (enumeration.IsPublic)
+                {
+                    sb.Append("    ").AppendLine($"/// <summary>A memory-optimized alternative to regular ToString() method on {enumeration.FullCsharpName} enum.</summary>");
+                }
 
                 sb.Append("    ").Append(methodVisibility).Append(" static string ToStringFast(this global::").Append(enumeration.FullCsharpName).AppendLine(" value)");
                 sb.AppendLine("    {");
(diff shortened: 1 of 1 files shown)


petrkoutnycz:
Output looks like this 
image

src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
```diff
@@ -146,9 +148,19 @@ private static string GenerateCode(ImmutableArray<EnumToProcess> enums)
                     sb.AppendLine("{");
                 }
 
+                if (typeIsPublic)
```
meziantou: I think you can add the comment on all types, even internal ones. This adds value as the xml comment is visible in the IDE.

src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
```diff
@@ -146,9 +148,19 @@ private static string GenerateCode(ImmutableArray<EnumToProcess> enums)
...
+                {
+                    sb.AppendLine($"/// <summary>A class with memory-optimized alternative to regular ToString() on enums.</summary>");
+                }
+
                 sb.Append(typeVisibility).AppendLine(" static partial class FastEnumToStringExtensions");
                 sb.AppendLine("{");
 
+                if (enumeration.IsPublic)
```
meziantou: Same here

src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
```diff
@@ -146,9 +148,19 @@ private static string GenerateCode(ImmutableArray<EnumToProcess> enums)
...
+                {
+                    sb.Append("    ").AppendLine($"/// <summary>A memory-optimized alternative to regular ToString() method on {enumeration.FullCsharpName} enum.</summary>");
```
meziantou: Would be nice to use a cref tag to link the enum. There is a class named DocumentationId (I'm not in front of a computer to check the name) that you can use to generate the cref value. If it's too complicated it's ok to not fix this comment.

meziantou (commented):
Thanks for the improvements!

src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
petrkoutnycz: I will check, thx

src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
petrkoutnycz: It depends as the comment does not add much value but rather satisfies the compiler 😃 though why not, can make it available for both public and non-public scenarios

src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
petrkoutnycz: Took a look and found `DocumentationCommenId` class but did not realize how to use it so I chose a naive approach. Let me know. It looks good though.

image

meziantou: I'll check your changes when I get back from vacation next week

petrkoutnycz: Thank you

petrkoutnycz:
@meziantou so what do you think? :-)

src/Meziantou.Framework.FastEnumToStringGenerator/EnumToStringSourceGenerator.cs
meziantou: I think you can add the `DocumentationId` property in the record `EnumToProcess`. BTW, you can replace the computed properties with `{get;} = ` to reduce some allocations.

````c#
    private sealed record EnumToProcess(ITypeSymbol EnumSymbol, List<EnumMemberToProcess> Members, bool IsPublic, string? Namespace)
    {
        public string FullCsharpName { get; } = EnumSymbol.ToString()!;
        public string? FullNamespace { get; } = Namespace ?? GetNamespace(EnumSymbol);
        public string? DocumentationId { get; } = DocumentationCommentId.CreateDeclarationId(EnumSymbol);

        // ...
    }
````

Then, you can use the new property in the `cref` attribute.

petrkoutnycz: Great, thank you, just updated the sources.

PR #557, opened 2023-09-15
Avoid extra allocation in MakePathRelativeTo. Fixes #556.

### LINKED ISSUE #556: Avoid extra string allocation in MakePathRelativeTo
The `MakePathRelativeTo` method currently allocates a new string - `rootPath._value + Path.DirectorySeparatorChar` - to pass to `PathDifference`. Since this is the only place that calls `PathDifference`, it is possible to avoid this allocation.

You can also remove the *"Same path except that path1 ended with a file name and path2 didn't"* branch, since `path1` is always assumed to be a directory.

```csharp
public string MakePathRelativeTo(FullPath rootPath)
{
    if (IsEmpty)
        throw new InvalidOperationException("The path is empty");

    if (rootPath.IsEmpty)
        return _value;

    if (rootPath == this)
        return ".";

    return PathDifference(rootPath._value, _value, compareCase: FullPathComparer.Default.IsCaseSensitive);
}

private static string PathDifference(string path1, string path2, bool compareCase)
{
    var directorySeparator = Path.DirectorySeparatorChar;

    int i;
    var si = -1;
    for (i = 0; (i <= path1.Length) && (i < path2.Length); ++i)
    {
        char c1 = i == path1.Length ? directorySeparator : path1[i];
        char c2 = path2[i];
       
        if ((c1 != c2) && (compareCase || (char.ToUpperInvariant(c1) != char.ToUpperInvariant(c2))))
            break;

        if (c1 == directorySeparator)
        {
            si = i;
        }
    }

    if (i == 0)
        return path2;

    if ((i == path1.Length + 1) && (i == path2.Length))
        return string.Empty;

    var relPath = new StringBuilder();
    // Walk down several dirs
    for (; i <= path1.Length; ++i)
    {
        char c = i == path1.Length ? directorySeparator : path1[i];
        if (c == directorySeparator)
        {
            relPath.Append("..");
            relPath.Append(directorySeparator);
        }
    }
    
    // NB: This can never be true:
    /*
    // Same path except that path1 ended with a file name and path2 didn't
    if (relPath.Length == 0 && path2.Length - 1 == si)
        return "." + directorySeparator; // Truncate the file name
    */
    
    return relPath.Append(path2.AsSpan(si + 1)).ToString();
}
```

RichardD2:
Removes the allocation of an extra string - `rootPath._value + Path.DirectorySeparatorChar` - in `MakePathRelativeTo`.

Also removes the unreachable *"Same path except that path1 ended with a file name and path2 didn't"* block.

Fixes #556.

### DIFF
--- a/src/Meziantou.Framework.FullPath/FullPath.cs
+++ b/src/Meziantou.Framework.FullPath/FullPath.cs
@@ -84,7 +84,7 @@ public readonly struct FullPath : IEquatable<FullPath>, IComparable<FullPath>
         if (rootPath == this)
             return ".";
 
-        return PathDifference(rootPath._value + Path.DirectorySeparatorChar, _value, compareCase: FullPathComparer.Default.IsCaseSensitive);
+        return PathDifference(rootPath._value, _value, compareCase: FullPathComparer.Default.IsCaseSensitive);
     }
 
     private static string PathDifference(string path1, string path2, bool compareCase)
@@ -93,9 +93,12 @@ public readonly struct FullPath : IEquatable<FullPath>, IComparable<FullPath>
 
         int i;
         var si = -1;
-        for (i = 0; (i < path1.Length) && (i < path2.Length); ++i)
+        for (i = 0; (i <= path1.Length) && (i < path2.Length); ++i)
         {
-            if ((path1[i] != path2[i]) && (compareCase || (char.ToUpperInvariant(path1[i]) != char.ToUpperInvariant(path2[i]))))
+            char c1 = i == path1.Length ? directorySeparator : path1[i];
+            char c2 = path2[i];
+
+            if ((c1 != c2) && (compareCase || (char.ToUpperInvariant(c1) != char.ToUpperInvariant(c2))))
                 break;
 
             if (path1[i] == directorySeparator)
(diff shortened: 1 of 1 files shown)


meziantou:
Some tests are failing with your changes in `Meziantou.Framework.Globbing`. For instance: `Meziantou.Framework.Globbing.Tests.GlobTests.GlobCollection1` and `Meziantou.Framework.Globbing.Tests.GlobTests.EnumerateFolder1`

RichardD2:
@meziantou Whoops! I missed replacing one instance of `path1[i]` with `c` in the original loop, which lead to an `IndexOutOfRangeException`.

That should be fixed now.

*(It doesn't help that I had to change the .NET 7 SDK in `global.json` from `302` to `401` to get the project to load properly, and comment out the `npm` commands since I don't have Node installed, before I could run the tests. Don't worry: I haven't committed those changes!)*

meziantou:
> It doesn't help that I had to change the .NET 7 SDK in global.json from 302 to 401 to get the project to load properly

The global.json should use `latestFeature` instead of `patch`. You can submit another PR if you want, or I'll do it when I switch to .NET 8 later this year.

PR #567, opened 2023-11-06
Fix context sensitive tests

0xced:
#### Fix TypeConverter TryConvert_DateTimeToDateTimeOffset test

It would fail with the following error when the current culture has a sensible day / month / year ordering.

> Expected value to represent the same point in time as `2018-06-12 -5h`, but `2018-12-06 -5h` does not.

#### Fix HumanReadableSerializer DateTime_Local test

Get the offset from the actual date, not from the current date since the offset might change when adjusted for daylight saving time.

Today, the test would fail with the following error.

```
Assert.Equal() Failure: Strings differ
                                ↓ (pos 21)
Expected: "2123-04-05T06:07:08+01:00"
Actual:   "2123-04-05T06:07:08+02:00"
                                ↑ (pos 21)
```

### Fix InlineSnapshotTesting prompt test

Trimming the milliseconds was not actually trimming the milliseconds but taking only the ticks per milliseconds part. The test would fail with the following error.

> The UTC time represented when the offset is applied must be between year 0 and 10,000.
> Parameter name: offset
>    at System.DateTimeOffset.ValidateDate(DateTime dateTime, TimeSpan offset)
>    at System.DateTimeOffset..ctor(Int64 ticks, TimeSpan offset)
>    at Meziantou.Framework.InlineSnapshotTesting.SnapshotUpdateStrategies.ProcessInfo.GetContextProcess() in Meziantou.Framework\src\Meziantou.Framework.InlineSnapshotTesting\SnapshotUpdateStrategies\ProcessInfo.Get.cs:line 31

### DIFF
--- a/tests/Meziantou.Framework.TypeConverter.Tests/DefaultConverterTests_DateTimeTo.cs
+++ b/tests/Meziantou.Framework.TypeConverter.Tests/DefaultConverterTests_DateTimeTo.cs
@@ -10,7 +10,7 @@ public class DefaultConverterTests_DateTimeTo
     public void TryConvert_DateTimeToDateTimeOffset()
     {
         var converter = new DefaultConverter();
-        var cultureInfo = CultureInfo.CurrentCulture;
+        var cultureInfo = new CultureInfo("en-US");
         var converted = converter.TryChangeType("6/12/2018 12:00:00 AM -05:00", cultureInfo, out DateTimeOffset value);
 
         converted.Should().BeTrue();
--- a/tests/Meziantou.Framework.TypeConverter.Tests/DefaultConverterTests_DateTimeTo.cs
+++ b/tests/Meziantou.Framework.TypeConverter.Tests/DefaultConverterTests_DateTimeTo.cs
@@ -10,7 +10,7 @@ public class DefaultConverterTests_DateTimeTo
     public void TryConvert_DateTimeToDateTimeOffset()
     {
         var converter = new DefaultConverter();
-        var cultureInfo = new CultureInfo("en-US");
+        var cultureInfo = CultureInfo.GetCultureInfo("en-US");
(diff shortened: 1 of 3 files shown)


tests/Meziantou.Framework.TypeConverter.Tests/DefaultConverterTests_DateTimeTo.cs
```diff
@@ -10,7 +10,7 @@ public class DefaultConverterTests_DateTimeTo
     public void TryConvert_DateTimeToDateTimeOffset()
     {
         var converter = new DefaultConverter();
-        var cultureInfo = CultureInfo.CurrentCulture;
+        var cultureInfo = new CultureInfo("en-US");
```
meziantou: Using `new CultureInfo(name)` returns a new editable instance. The static method uses another ctor which returns a readonly instance. It also caches the instances. If you are curious, you can check [the implementation](https://github.com/dotnet/runtime/blob/e733539a38458c5c3a64688b55a702444e470700/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.cs#L1054)

```suggestion
        var cultureInfo = CultureInfo.GetCultureInfo("en-US");
```

0xced: Good point, I didn't know! Will have some cleanup to perform in my own code. 😉

meziantou:
Thanks for the fix 🚀
Do you want me to publish a new version soon, or can I wait for .NET 8 (in 2 weeks)?

0xced:
I'm not in a hurry, and this pull request only fixes tests anyway! 😆

I have more pull requests coming and I plan to build on top of #568 to improve the ResX source generator. I plan to add an option to generate a strongly typed resource class with non static members. And #568 is just the first step in that direction.

meziantou:
Can you open an issue to explain your plan and be sure it will be accepted before you start the implementation?

PR #568, opened 2023-11-06
Use Verify.Xunit for ResxGenerator tests

0xced:
Comparing the whole generated strongly typed resource class instead of some parts of it will become handy when modifying the Resx source generator in later pull requests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,9 @@ bld/
 *.VisualState.xml
 TestResult.xml
 
+# Verify
+*.received.*
+
 # Build Results of an ATL Project
 [Dd]ebugPS/
 [Rr]eleasePS/
--- a/tests/Directory.Build.props
+++ b/tests/Directory.Build.props
@@ -6,10 +6,10 @@
   </ItemGroup>
 
   <ItemGroup Condition="$(TargetFramework) != 'netstandard2.0'">
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.6.0" />
-    <PackageReference Include="FluentAssertions" Version="6.11.0" />
-    <PackageReference Include="xunit" Version="2.5.0" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.0">
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
+    <PackageReference Include="FluentAssertions" Version="6.12.0" />
+    <PackageReference Include="xunit" Version="2.6.1" />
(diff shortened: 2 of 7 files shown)


meziantou:
There are two test projects for this source generator. 
- The one you edited validates the source generator behavior (read configuration, output errors if needed, handle invalid resx, etc.)
- `Meziantou.Framework.ResxSourceGenerator.GeneratorTests` validates the generated code behaves correctly by executing the generated code

Testing the generated code as text seems like testing an implementation detail. You can assert invalid code and have a green test. What you want is to validate the generated code behaves correctly and the public API is usable as expected. That's what `ResxSourceGenerator.GeneratorTests` does.

Also, validating the full source generator output obfuscates what is actually being tested. I understand it's nice to view the changed file when you edit the source generator, but I don't think it provides lot of value.

So, I'm not sure I'm ok with this change.

0xced:
That's a very good point. Unfortunately I'm having trouble testing the actual source generator with the `Meziantou.Framework.ResxSourceGenerator.GeneratorTests` project.

Setting a class name, for example, has no effect.

```xml
<ItemGroup>
  <ProjectReference Include="..\..\src\Meziantou.Framework.ResxSourceGenerator\Meziantou.Framework.ResxSourceGenerator.csproj" PrivateAssets="all" ReferenceOutputAssembly="false" OutputItemType="Analyzer" />
  <ProjectReference Include="..\TestUtilities\TestUtilities.csproj" />
  <AdditionalFiles Include="**/*.resx" ClassName="ItIsIgnored" />
</ItemGroup>
```

All `ResxGeneratorTests` are still passing even after changing the class name to `ItIsIgnored`, and without adapting the test. It should not even compile!

I have successfully managed to run tests against a built NuGet package in the `Serilog.Settings.Configuration` project, but it was a [huge pain](https://github.com/serilog/serilog-settings-configuration/blob/1471d91d9bb29cc37892ae8e51847d5bc13f5dca/test/Serilog.Settings.Configuration.Tests/Support/TestApp.cs).

Do you have an idea how the options of the source generator could be tested, ideally without actually building the NuGet package?

meziantou:
I think the props `Meziantou.Framework.ResxSourceGenerator.props` must be imported in the project to get the same behavior as the package.

Also, Visual Studio gives a bad experience when referencing a source generator project. You should run the tests using a CLI (dotnet test). IIRC I think you must the `--disable-build-servers` options with the latest SDK to be sure to reload the source generator each time you run the test. If the option is not available it's not needed.

PR #577, opened 2023-11-17
Add support for generic attributes to StronglyTypedIdAttribute

rfvgyhn:
This PR adds support for generic attributes to StronglyTypedIdAttribute. `[StronglyTypedId<int>]`

I wasn't sure what the best way to conditionally add a generic version of the attribute. This approach adds an MSBuild file that defines a preprocessor constant to check if the project supports generic attributes.


meziantou:
Can you rebase your branch? Also, I changed the way attributes are defined. It's now in the Meziantou.Framework.StronglyTypedId.Annotations project. You should be able to use `#if NET8_0_OR_GREATER` to define the attribute.

rfvgyhn:
Generic attributes are a csharp 11/.net 7 feature. Do you still want to gate it behind .net8?

meziantou:
Use the minimum supported version. If it's .net 7, use NET_7_0_OR_GREATER.

rfvgyhn:
I haven't worked with multi-targeted libs before so correct me if I'm wrong on the following.

Wrapping the attribute definition in `NET7_0_OR_GREATER` means the reference to the Annotations project in the test project needs to get rid of the `TargetFramework=netstandard2.0` property since `NET7_0_OR_GREATER` would always be false. That then means adding nuget references to the compilation object needs to also account for specific Microsoft.NETCore.App.Ref versions. Does that sound right?

https://github.com/meziantou/Meziantou.Framework/blob/3f7bd16f1924b541fd81fe56271b6563d9b3e9a0/tests/Meziantou.Framework.StronglyTypedId.Tests/Meziantou.Framework.StronglyTypedId.Tests.csproj#L9

https://github.com/meziantou/Meziantou.Framework/blob/3f7bd16f1924b541fd81fe56271b6563d9b3e9a0/tests/Meziantou.Framework.StronglyTypedId.Tests/StronglyTypedIdSourceGeneratorTests.cs#L40-L46

```diff
+ const string NetCoreVersion =
+ #if NET7_0
+         "7.0.14"
+ #elif NET8_0
+         "8.0.0"
+ #else
+         "6.0.12"
+ #endif
+         ;
var compilation = await CreateCompilation(sourceText,
        [
-            new NuGetReference("Microsoft.NETCore.App.Ref", "6.0.12", "ref/"),
+            new NuGetReference("Microsoft.NETCore.App.Ref", NetCoreVersion, "ref/"),
```

meziantou:
Seems ok. I wonder if you can only reference the .NET 8 package in the test or if you need to use `#if`.

rfvgyhn:
Only referencing .net8 would produce `CS1701` errors for most tests.

```
Expected result.Success to be true because Project cannot build:
error CS1701: Assuming assembly reference 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' used by 'Meziantou.Framework.StronglyTypedId.Annotations' matches identity 'System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' of 'System.Runtime', you may need to supply runtime policy
```

I've rebased the branch. Let me know what you think.

PR #595, opened 2024-02-14
Add support for domain credential type in Meziantou.Framework.Win32.CredentialManager

nixtar:
Hello,

I needed the ability to create `CRED_TYPE_DOMAIN_PASSWORD` entries in the credential manager from a c# desktop application.

I found some other credential manager dotnet libs out there that exposes the type. However, Meziantou.Framework.Win32.CredentialManager has a much nicer API so I decided to add support for specifying the CredentialType for the following methods:

ReadCredential
WriteCredential
DeleteCredential

For DeleteCredential I limited it to only CredentialType.Generic and CredentialType.DomainPassword as this is all I have the ability to test etc. `CRED_TYPE_DOMAIN_VISIBLE_PASSWORD` ( `DomainVisiblePassword` ) might be able to be supported but I was wary when I saw `This value is no longer supported` in the docs https://learn.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credentiala.

Also added tests.

Please let me know if you have any issues with these changes that you would like me to address. 😄

 
Thanks,
Nick

### DIFF
--- a/src/Meziantou.Framework.Win32.CredentialManager/CredentialManager.cs
+++ b/src/Meziantou.Framework.Win32.CredentialManager/CredentialManager.cs
@@ -48,12 +48,12 @@ public static class CredentialManager
         return new Credential((CredentialType)credential->Type, applicationName, userName, secret, comment);
     }
 
-    public static void WriteCredential(string applicationName, string userName, string secret, CredentialPersistence persistence)
+    public static void WriteCredential(string applicationName, string userName, string secret, CredentialPersistence persistence, CredentialType type = CredentialType.Generic)
     {
-        WriteCredential(applicationName, userName, secret, comment: null, persistence);
+        WriteCredential(applicationName, userName, secret, comment: null, persistence, type);
     }
 
-    public static unsafe void WriteCredential(string applicationName, string userName, string secret, string? comment, CredentialPersistence persistence)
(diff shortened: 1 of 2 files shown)


src/Meziantou.Framework.Win32.CredentialManager/CredentialManager.cs
```diff
@@ -14,9 +14,9 @@ namespace Meziantou.Framework.Win32;
 [SupportedOSPlatform("windows5.1.2600")]
 public static class CredentialManager
 {
-    public static unsafe Credential? ReadCredential(string applicationName)
+    public static unsafe Credential? ReadCredential(string applicationName, CredentialType type = CredentialType.Generic)
```
meziantou: Adding an optional parameter is a binary breaking change. You should preserve the original method and add a new overload. (same comment for other new methods)

meziantou (changes requested):

src/Meziantou.Framework.Win32.CredentialManager/CredentialManager.cs
nixtar: Yes of course! Sorry about that.
I've restored the original method signatures with additional overloads.

Tested by compiling the unit test project from the main branch then replaced the `Meziantou.Framework.Win32.CredentialManager.dll` manually from ones compiled with the latest commit.

meziantou (approved):

PR #614, opened 2024-05-30
Added functions to get and disable the CPU rate hard cap.

### LINKED ISSUE #613: `Get` functions for JobObject
I need to be able to read values from a job object.  For example, I need `GetCpuRateHardCap()`.  Are you open to a pull request if I add the functions I need?   CONTRIBUTING.md mentions pull requests for bug fixed, but not for new features.

Thanks.

toschlog-sony:
Added two functions: GetCpuRateHardCap() and DisableCpuRateHardCap().
Added tests for new functions.
Fixed misspelling of "weight".

fix #613

### DIFF
--- a/src/Meziantou.Framework.Win32.Jobs/JobObject.cs
+++ b/src/Meziantou.Framework.Win32.Jobs/JobObject.cs
@@ -196,20 +196,52 @@ public sealed class JobObject : IDisposable
     }
 
     /// <summary>
+    /// Get the job's CPU rate limit enabled status and value.
+    /// </summary>
+    /// <returns>Bool indicating if CPU rate control is enabled and the job's CPU rate limit.</returns>
+    public unsafe (bool, int) GetCpuRateHardCap()
+    {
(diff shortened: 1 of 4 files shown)


src/Meziantou.Framework.Win32.Jobs/JobObject.cs
```diff
@@ -195,21 +195,53 @@ public unsafe void SetCpuRateHardCap(int cpuRate)
         }
     }
 
+    /// <summary>
+    /// Get the job's CPU rate limit enabled status and value.
+    /// </summary>
+    /// <returns>Bool indicating if CPU rate control is enabled and the job's CPU rate limit.</returns>
+    public unsafe (bool, int) GetCpuRateHardCap()
```
meziantou: I prefer returning a class or a struct instead of a ValueTuple. This is easier to extend later without introducing breaking changes. Note that I do not care for internal methods, only the public ones.

toschlog-sony:
Now returning a struct.

PR #623, opened 2024-06-10
Added TryOpen()

toschlog-sony:
We're using the framework for an app that polls twice a second for JobObjects.  Those objects don't usually exist, so a lot of exceptions are being thrown.  I've added a `TryOpen()` function to use instead.


src/Meziantou.Framework.Win32.Jobs/JobObject.cs
```diff
@@ -75,6 +75,28 @@ public static JobObject Open(JobObjectAccessRights desiredAccess, bool inherited
         return new JobObject(handle);
     }
 
+    // Win32 error code
+    private const int ERROR_FILE_NOT_FOUND = 2;
```
meziantou: I think you can move this to https://github.com/meziantou/Meziantou.Framework/blob/main/src/Meziantou.Framework.Win32.Jobs/NativeMethods.txt. The source generator will generate the constant name and value.

toschlog-sony:
Looks like there's some sort of API key problem deploying the packages.

meziantou:
> Looks like there's some sort of API key problem deploying the packages.

Secrets are not available for forks. But, I can merge without it :)

Thanks for your contribution 🚀

PR #664, opened 2024-10-17
Add parsing new section headings in CODEOWNERS files

Totoo98:
Hi! My team would like to be able to parse new section headings in CODEOWNERS files, namely the required reviewer count and the default owners (https://docs.gitlab.com/ee/user/project/codeowners/reference.html#section-headings).

This change improves the parsing of section lines, and adds a few tests for section heading parsing as well as a test for files with lines ending with CRLF.

Please note there is a breaking change as the CodeOwnersSection constructor now takes an int instead of bool, since we can now deduce IsOptional from RequiredReviewerCount.
I couldn't find a breaking change guidelines for this project, but if we don't want one right now I suggest keeping both constructors, set the RequiredReviewerCount at 1 if isOptional is false in the old one, and ideally mark it as deprecated.

Thanks ⁠😊
Thomas

### DIFF
--- a/src/Meziantou.Framework.CodeOwners/CodeOwnersSection.cs
+++ b/src/Meziantou.Framework.CodeOwners/CodeOwnersSection.cs
@@ -5,14 +5,21 @@ namespace Meziantou.Framework.CodeOwners;
 [StructLayout(LayoutKind.Auto)]
 public readonly struct CodeOwnersSection : IEquatable<CodeOwnersSection>
 {
-    public CodeOwnersSection(string name, bool isOptional)
+    public CodeOwnersSection(string name, int requiredReviewerCount = 1, IReadOnlyCollection<string> defaultOwners = null)
     {
         Name = name;
-        IsOptional = isOptional;
+        RequiredReviewerCount = requiredReviewerCount;
+        DefaultOwners = defaultOwners ?? new List<string>();
     }
 
     public string Name { get; }
-    public bool IsOptional { get; }
+
+    public int RequiredReviewerCount { get; }
+    public bool IsOptional => RequiredReviewerCount == 0;
+    public bool IsMandatory => !IsOptional;
+
+    public IReadOnlyCollection<string> DefaultOwners { get; }
+    public bool HasDefaultOwners => DefaultOwners is not null && DefaultOwners.Count > 0;
 
     public override bool Equals(object? obj)
     {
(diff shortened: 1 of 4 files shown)


src/Meziantou.Framework.CodeOwners/CodeOwnersSection.cs
```diff
@@ -5,14 +5,21 @@ namespace Meziantou.Framework.CodeOwners;
 [StructLayout(LayoutKind.Auto)]
 public readonly struct CodeOwnersSection : IEquatable<CodeOwnersSection>
 {
-    public CodeOwnersSection(string name, bool isOptional)
+    public CodeOwnersSection(string name, int requiredReviewerCount = 1, IReadOnlyCollection<string> defaultOwners = null)
```
meziantou: Can you make it `internal`? I don't think it need to be public.

tests/Meziantou.Framework.CodeOwners.Tests/CodeOwnersParserTests.cs
```diff
@@ -180,4 +180,86 @@ public void ParseCodeOwnersWithPatternsWithoutMembers()
...
+
+        var expected = new CodeOwnersEntry[]
+        {
+            CodeOwnersEntry.FromUsername(0, "*", "user1", section: new CodeOwnersSection("Test1")),
+            CodeOwnersEntry.FromUsername(1, "*", "user2", section: new CodeOwnersSection("Test2", 2)),
+            CodeOwnersEntry.FromUsername(2, "*", "user3", section: new CodeOwnersSection("Test3", 1, ["@defaultOwner"])),
+        };
+
+        actual.Should().Equal(expected);
+    }
```
meziantou, line 374: What's the behavior of GitLab in the following cases? Can you add test that matches the GitLab bahvior?
- `^[section][2]` (From the doc "Optional sections ignore the number of approvals required")
- `[section] [2]`
- `[section] # [2]`

meziantou:
Can you bump the version in the csproj, so it will deploy a new version after merging this PR?

src/Meziantou.Framework.CodeOwners/CodeOwnersSection.cs
Totoo98: Sure, addressed 👌

tests/Meziantou.Framework.CodeOwners.Tests/CodeOwnersParserTests.cs
Totoo98: Good catch, I'll have a look next week into those edge cases :)

Totoo98: I've updated the parser and tests according to how GitLab handle those cases, as well as a few more similar ambiguities.

PR #667, opened 2024-10-21
Add option to present the login failure error on the credentials prompt

jonasdmentia:
Hopefully this is obvious enough, just adding in minimal support for presenting a login failure error on the credential's prompt (i.e. the user just tried and invalid username/password combination so the prompt is displayed again, this time with the error message). It'll be easy enough to expand to include other errors if that makes sense, but for now it just exposes the single error ERROR_LOGON_FAILURE.


meziantou:
Thanks for your contribution 🚀

PR #679, opened 2024-11-16
Fix TaskEx.WhenAll formatting

0xced:
Nothing fancy, just better alignment of generated code from the T4 text template.

Note that I did it with Rider so it would be worth trying to generate it again with Visual Studio and ensure that there are no differences.


meziantou (approved):

meziantou:
Looks good on VS too :)

PR #696, opened 2024-12-16
Update readme.md

STRUDSO:
Changed  scrub xml and scrub json to serializer and moved them up together with the other serializer snippets.

Also a question was trying to play around with this for mvc problem details, got a traceid as part of output. 
I could perhaps do a regex to match a traceid, but would rather not...
I also tried to just manipulate the object like so:
````c#
 var content = await result.Content.ReadFromJsonAsync<ProblemDetails>();

            Assert.NotNull(content);
            content.Extensions["traceId"] = "[any trace...]";

            InlineSnapshot
                .Validate
````

so my solution ended being this

````c#

 var content = await result.Content.ReadFromJsonAsync<ProblemDetails>();
            InlineSnapshot
                .WithSettings(x => x.ScrubLinesWithReplace(
                    x => x.Contains("traceid", StringComparison.OrdinalIgnoreCase) ? "  traceId: [any trace...]" : x
                    ))
                .Validate(content, """
                    Title: Validation Error
                    Status: 400
                    Detail:
                      Validation failed: 
                       -- PageNumber: 'Page Number' must be greater than '0'. Severity: Error
                    Instance: GET /meetinggroupproposals
                    Extensions:
                      traceId: [any trace...]
                      Errors:
                        [
                          {
                            "key": "PageNumber",
                            "value": "'Page Number' must be greater than '0'."
                          }
                        ]
                    """);
        }
                    
````

but maybe there is a better way?

### DIFF
--- a/src/Meziantou.Framework.InlineSnapshotTesting/readme.md
+++ b/src/Meziantou.Framework.InlineSnapshotTesting/readme.md
@@ -274,6 +274,32 @@ InlineSnapshot
 ````
 
 ````c#
+var data = JsonNode.Parse("""{ "prop": "value" }""");
+InlineSnapshot
+    .WithSerializer(options => options.ScrubJsonValue("$.prop", node => "[redacted]"))
+    .Validate(data, """
+        {
+          "prop": "[redacted]"
+        }
+        """);
+````
+
+````c#
+var data = XDocument.Parse("""
+                <root>
+                  <item attr="1">test1</item>
+                </root>
+                """);
+InlineSnapshot
+    .WithSerializer(options => options.ScrubXmlAttribute("//item/@attr", attribute => "[redacted]"))
+    .Validate(data, """
+        <root>
+          <item attr="[redacted]">test1</item>
+        </root>
+        """);
+````
+
+````c#
 InlineSnapshot
     .WithSettings(settings => settings.ScrubLines(line => line.Contains("dummy")))
     .Validate("abc\ndummy", "abc");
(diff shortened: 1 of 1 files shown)


meziantou:
> but maybe there is a better way?

You can use a custom converter

```c#
var converter = new ScrubDictionaryKeyValuePairValue((key, value) =>
{
    if (key == "traceId")
        return "[redacted]";
    return value;
});
options.AddAttribute<ProblemDetails>(obj => obj.Extensions, new HumanReadableConverterAttribute(converter));
````
````c#
class ScrubDictionaryKeyValuePairValue(Func<string, object?, object?> scrubber) : HumanReadableConverter<IEnumerable<KeyValuePair<string, object?>>>
{
    protected override void WriteValue(HumanReadableTextWriter writer, IEnumerable<KeyValuePair<string, object?>> value, HumanReadableSerializerOptions options)
    {
        Debug.Assert(value is not null);
        if (options.DictionaryKeyOrder is not null)
        {
            value = value.OrderBy(value => value.Key, options.DictionaryKeyOrder);
        }

        var hasItem = false;
        foreach (var prop in value)
        {
            var scrub = scrubber(prop.Key, prop.Value);
            if (scrub is null)
                continue;

            if (!hasItem)
            {
                writer.StartObject();
                hasItem = true;
            }

            writer.WritePropertyName(prop.Key);
            HumanReadableSerializer.Serialize(writer, scrub, options);
        }

        if (hasItem)
        {
            writer.EndObject();
        }
        else
        {
            writer.WriteEmptyObject();
        }
    }
}
````

meziantou (approved):

PR #734, opened 2025-02-20
Add Spanish l10n for relative dates

zalito12:
Hi! There is no much info about naming on contribution, but I would like to add the spanish translations for relative dates 😄


tests/Meziantou.Framework.ResxSourceGenerator.GeneratorTests/Resource1.es.resx
```diff
@@ -0,0 +1,27 @@
+<root>
```
meziantou: Can you remove this file as it is not needed for the RelativeDate?

zalito12: Sure, I will. It was created automatically, actually I didn't know its purpose 😆

PR #813, opened 2025-06-20
Update Meziantou.Framework.HttpClientMock readme.md

Meir017:


src/Meziantou.Framework.HttpClientMock/readme.md
```diff
@@ -85,10 +85,10 @@ You can forward logs to the xUnit `ITestOutputHelper`. This can be useful to deb
 1. Add the `Meziantou.Extensions.Logging.Xunit` package to your project
 1.
 
-    ````c#
-    using var loggerProvider = new XUnitLoggerProvider(testOutputHelper);
-    await using var mock = new HttpClientMock(loggerProvider);
-    ````
+````c#
```
meziantou: That's not the right fix for the formatting. The code block should be inside the list item, so it needs to be indented. However, there is an extra blank line before the code block, so it's doesn't work as intended. You can keep the indentation and remove the extra line (line 87) instead.

meziantou:
> Before this change the code snippet is not being rendered as c#

I agree and this needs to be fixed. I've updated your branch to reflect my comment. You can see that the code block is inside the `<li>` element now.

image

Thanks for reporting the issue 🚀

PR #845, opened 2025-07-20
Use microsoft testing platform runner

meziantou:


Frulfump:
Maybe you have seen  this already but in case you haven't https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test?tabs=dotnet-test-with-mtp is that something you would consider using here?

Frulfump:
Was this closed due to the removal of dotnet.config in .NET 10?

Issue #939, opened 2025-10-17
ResxSourceGenerator with linked file from relative path generates empty file


fowl2:
Hi there, just trying out this source generator. Works when the resx file is directly under the project, but when I use a relative path to a file outside thee project folder, the generated file is empty.

My project file:

D:\source\MyProject\src\MyProjectPlugin\MyProjectPlugin\MyProjectPlugin.csproj

``` xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFrameworks>net462;net9.0</TargetFrameworks>
  </PropertyGroup>

  <PropertyGroup>
    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
    <CompilerGeneratedFilesOutputPath>$(BaseIntermediateOutputPath)\GeneratedFiles</CompilerGeneratedFilesOutputPath>
    <DefaultResourcesNamespace>MyProjectPluginResources</DefaultResourcesNamespace>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Meziantou.Framework.ResxSourceGenerator">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
  </ItemGroup>

  <ItemGroup>
    <AdditionalFiles Include="..\..\script\Translation\localizedString.1033.resx" Link="Properties\localizedString.1033.resx" />
  </ItemGroup>

</Project>

```

And the generated file:

D:\source\MyProject\src\MyProjectPlugin\MyProjectPlugin\obj\GeneratedFiles\Meziantou.Framework.ResxSourceGenerator\Meziantou.Framework.ResxSourceGenerator.ResxGeneratorlocalizedString.1033.resx.g.cs:
``` csharp

// Debug info:
// key: D:\source\MyProject\src\MyProjectPlugin\MyProjectPlugin\..\..\script\Translation\localizedString.1033
// files: D:\source\MyProject\src\MyProjectPlugin\MyProjectPlugin\..\..\script\Translation\localizedString.1033.resx
// RootNamespace (metadata): MyProjectPlugin
// ProjectDir (metadata): D:\source\MyProject\src\MyProjectPlugin\MyProjectPlugin\
// Namespace / DefaultResourcesNamespace (metadata): MyProjectPluginResources
// ResourceName (metadata): 
// ClassName (metadata): 
[...]

PR #940, opened 2025-10-17
Detect resource name from EmbeddedResource

meziantou:
Fix https://github.com/meziantou/Meziantou.Framework/issues/939


fowl2:
Hi not sure if I understand this correctly but note that my resources aren't actually embedded in the assembly where the generator is running - I just want the key name strings. Will this still work?

meziantou:
Yes, it can work. However, with the current version you may need to specify a resource name manually (can be whatever you want as it won't be used)

```
<AditionalFiles Include="sample.resx" ResourceName="dummy" GenerateKeyNamesType="false" />
```

`GenerateKeyNamesType` disable generating the class with property to access the resx file. The KeyNames class will still be generated.

PR #1270, opened 2026-04-21
Use Meziantou.Framework.TextDiff in InlineSnapshotTesting

meziantou:
`InlineDiffAssertionMessageFormatter` contained a full ~280-line custom Myers diff implementation that duplicated the logic already available in `Meziantou.Framework.TextDiff`. This PR removes that duplication by delegating to `TextDiff.ComputeDiff()`.

## Changes

**`Meziantou.Framework.InlineSnapshotTesting`**
- Added a `ProjectReference` to `Meziantou.Framework.TextDiff`
- Replaced the entire custom diff implementation in `InlineDiffAssertionMessageFormatter` with a simple `TextDiff.ComputeDiff()` call (~280 lines removed, ~25 added)

**`Meziantou.Framework.TextDiff`** (extended to support net472/net48, which InlineSnapshotTesting requires)
- Added `net472` and `net48` to `TargetFrameworks`
- Added `System.Memory` package reference for those TFMs
- `TextChunker.cs`: Guarded `SearchValues<T>` (a .NET 8+ API) with `#if NET8_0_OR_GREATER`, falling back to `char[]` + `Array.IndexOf`
- `TextDiff.cs`: Guarded `string.GetHashCode(span, comparison)` with `#if NET5_0_OR_GREATER`
- `PatienceDiff.cs`, `HistogramDiff.cs`, `HuntSzymanskiDiff.cs`: Replaced `Array.Fill` (unavailable on net4x) with explicit initialization loops

All existing `InlineDiffAssertionMessageFormatterTests` (7 tests) and `TextDiffTests` (84 tests) pass.

### DIFF
--- a/src/Meziantou.Framework.TextDiff/TextChunker.cs
+++ b/src/Meziantou.Framework.TextDiff/TextChunker.cs
@@ -1,4 +1,6 @@
+#if NET8_0_OR_GREATER
 using System.Buffers;
+#endif
 
 namespace Meziantou.Framework;
 
@@ -13,7 +15,11 @@ public class TextChunker
 
     private sealed class LineChunker : TextChunker
     {
+#if NET8_0_OR_GREATER
         private static SearchValues<char> NewLineCharacters { get; } = SearchValues.Create("\r\n\u0085\u2028\u2029");
+#else
+        private static readonly char[] NewLineCharacters = ['\r', '\n', '\u0085', '\u2028', '\u2029'];
+#endif
 
         public override IEnumerable<string> Chunk(ReadOnlySpan<char> value)
         {
@@ -47,7 +53,11 @@ public class TextChunker
 
     private sealed class WordChunker : TextChunker
     {
+#if NET8_0_OR_GREATER
         private static SearchValues<char> WhiteSpaceCharacters { get; } = SearchValues.Create("\t\n\v\f\r\u0020\u0085\u00A0\u1680\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u2028\u2029\u202F\u205F\u3000");
+#else
+        private static readonly char[] WhiteSpaceCharacters = ['\t', '\n', '\v', '\f', '\r', '\u0020', '\u0085', '\u00A0', '\u1680', '\u2000', '\u2001', '\u2002', '\u2003', '\u2004', '\u2005', '\u2006', '\u2007', '\u2008', '\u2009', '\u200A', '\u2028', '\u2029', '\u202F', '\u205F', '\u3000'];
+#endif
 
         public override IEnumerable<string> Chunk(ReadOnlySpan<char> value)
         {
@@ -69,7 +79,11 @@ public class TextChunker
                 }
 
                 var whiteSpaceEnd = whiteSpaceStart + 1;
+#if NET8_0_OR_GREATER
                 while (whiteSpaceEnd < value.Length && WhiteSpaceCharacters.Contains(value[whiteSpaceEnd]))
+#else
+                while (whiteSpaceEnd < value.Length && Array.IndexOf(WhiteSpaceCharacters, value[whiteSpaceEnd]) >= 0)
+#endif
                 {
                     whiteSpaceEnd++;
                 }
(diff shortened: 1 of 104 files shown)


src/Meziantou.Framework.TextDiff/TextDiff.cs
```diff
@@ -104,7 +104,15 @@ public bool Equals(string? x, string? y)
         public int GetHashCode(string obj)
         {
             ArgumentNullException.ThrowIfNull(obj);
-            return string.GetHashCode(Trim(obj.AsSpan()), _comparison);
+            var trimmed = Trim(obj.AsSpan());
+#if NET5_0_OR_GREATER
```
meziantou: Use new version of Meziantou.polyfill 1.0.113 so this `#if` is not needed

src/Meziantou.Framework.TextDiff/TextChunker.cs
```diff
@@ -13,7 +15,11 @@ public virtual IEnumerable<string> Chunk(ReadOnlySpan<char> value)
 
     private sealed class LineChunker : TextChunker
     {
+#if NET8_0_OR_GREATER
```
meziantou: I think the `#if` is not needed as there is a polyfill for search values

meziantou: Accepted. I removed the `#if` branches in `TextChunker` and now use `SearchValues<char>` for all targets with Polyfill 1.0.113. I also switched to a `SearchValues.Contains` based index helper so it stays cross-target compatible.

src/Meziantou.Framework.TextDiff/TextDiff.cs
meziantou: Accepted. I updated `Meziantou.Polyfill` to `1.0.113` and removed the `#if` fallback in `TextDiff.GetHashCode`, so we now always use `string.GetHashCode(span, comparison)`.

src/Meziantou.Framework.TextDiff/TextChunker.cs
```diff
@@ -11,6 +11,17 @@ public class TextChunker
     public virtual IEnumerable<string> Chunk(ReadOnlySpan<char> value)
         => Lines.Chunk(value);
 
+    private static int IndexOfAny(ReadOnlySpan<char> value, SearchValues<char> searchValues)
```
meziantou: Use the latest version of Meziantou.Polyfill and remove this method. you can use

var lineEndIndex = value[start..].IndexOfAny(NewLineCharacters);

meziantou: Partially accepted. I removed the helper method and now use `value[start..].IndexOfAny(...)` directly in `TextChunker`. I did not bump Polyfill further in this change because the branch already uses 1.0.113 and this refactor is sufficient for the requested simplification.

src/Meziantou.Framework.TextDiff/TextChunker.cs
```diff
@@ -85,6 +85,7 @@ public override IEnumerable<string> Chunk(ReadOnlySpan<char> value)
 
             return words;
         }
+
```
meziantou: Remove the new line

src/Meziantou.Framework.TextDiff/TextChunker.cs
```diff
@@ -69,7 +69,7 @@ public override IEnumerable<string> Chunk(ReadOnlySpan<char> value)
                 }
 
                 var whiteSpaceEnd = whiteSpaceStart + 1;
-                while (whiteSpaceEnd < value.Length && WhiteSpaceCharacters.Contains(value[whiteSpaceEnd]))
+                while (whiteSpaceEnd < value.Length && value[whiteSpaceEnd..].IndexOfAny(WhiteSpaceCharacters) == 0)
                 {
                     whiteSpaceEnd++;
                 }
```
meziantou, line 80: I think you can use IndexOfAnyExcept here to avoid looping characters by character

You also need to update the Meziantou.Polyfill package to 1.0.116

meziantou: Accepted. I switched the whitespace run detection to `IndexOfAnyExcept(WhiteSpaceCharacters)` and updated `Meziantou.Polyfill` to `1.0.116`.

src/Meziantou.Framework.TextDiff/TextChunker.cs
meziantou: Accepted. I removed the extra blank line in `TextChunker.cs`.

PR #1331, opened 2026-04-28
Enable nullable in selected serializer-related test projects

meziantou:
## Why
Several targeted test projects explicitly disabled nullable reference type analysis. This change enables nullable in those projects and updates tests to satisfy compiler analysis while keeping the same test behavior.

## What changed
- Removed `<Nullable>disable</Nullable>` from:
  - `Meziantou.Framework.HumanReadableSerializer.Tests`
  - `Meziantou.Framework.InlineSnapshotTesting.Tests`
  - `Meziantou.Framework.JsonPath.Tests`
  - `Meziantou.Framework.MediaTags.Tests`
- Updated shared test helpers (`SerializerTestsBase.Validation` and `AssertSerialization` overloads) to accept nullable inputs.
- Adjusted affected tests and local test-only helper types with nullable annotations, null-forgiving where appropriate, and existing-behavior-preserving assertions.
- Kept test logic and expected snapshots/output unchanged.

## Validation
- `dotnet build` succeeds for the affected projects.
- `dotnet test` succeeds for:
  - `Meziantou.Framework.HumanReadableSerializer.Tests`
  - `Meziantou.Framework.InlineSnapshotTesting.Tests`
  - `Meziantou.Framework.JsonPath.Tests`
  - `Meziantou.Framework.MediaTags.Tests`
- Required scripts were run:
  - Succeeded: `dotnet run ./eng/update-trimmable.cs`
  - Failed due pre-existing repo-wide target framework mismatch checks unrelated to this change: `update-bom.cs`, `update-readme.cs`, `update-project-slnx.cs`, `validate-testprojects-configuration.cs`

### DIFF
--- a/tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTests.cs
+++ b/tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTests.cs
@@ -68,7 +68,7 @@ public sealed partial class SerializerTests : SerializerTestsBase
     public void SerializeObject_Null() => AssertSerialization(null, "<null>");
 
     [Fact]
-    public void SerializeObject_Null_Nested() => AssertSerialization(new { Obj = (object)null }, "Obj: <null>");
+    public void SerializeObject_Null_Nested() => AssertSerialization(new { Obj = (object?)null }, "Obj: <null>");
 
     [Fact]
     public void SerializeArray_Empty()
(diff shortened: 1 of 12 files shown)


tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTests.cs
```diff
@@ -2288,11 +2288,11 @@ public StructWithDefaultConstructor()
 
     private sealed class ObjectWithFields
     {
-        private string _privateField;
+        private string _privateField = null!;
```
meziantou:
```suggestion
        private string? _privateField;
```

tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTests.cs
```diff
@@ -2288,11 +2288,11 @@ public StructWithDefaultConstructor()
...
 
         [HumanReadableInclude]
-        private string _privateFieldIncluded;
-        public string FieldString;
+        private string _privateFieldIncluded = null!;
+        public string FieldString = null!;
```
meziantou:
```suggestion
        private string? _privateFieldIncluded;
        public string? FieldString;
```

tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTests.cs
```diff
@@ -2389,11 +2389,11 @@ private sealed class ClassWithAttributes
         public int Prop2 { get; set; }
 
         [HumanReadableConverter(typeof(CustomStringConverter))]
-        public string Prop3 { get; set; }
+        public string Prop3 { get; set; } = null!;
```
meziantou:
```suggestion
        public string? Prop3 { get; set; }
```

tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTestsBase.cs
```diff
@@ -5,23 +5,23 @@ public class SerializerTestsBase
 {
     protected sealed record Validation
     {
-        public object Subject { get; init; }
-        public string Expected { get; init; }
-        public Type Type { get; init; }
-        public HumanReadableSerializerOptions Options { get; init; }
+        public object? Subject { get; init; }
+        public string Expected { get; init; } = null!;
```
meziantou:
```suggestion
        public string? Expected { get; init; }
```

tests/Meziantou.Framework.JsonPath.Tests/JsonPathParseTests.cs
```diff
@@ -59,14 +59,14 @@ public void Parse_InvalidExpression(string expression)
     [Fact]
     public void TryParse_NullExpression_ReturnsFalse()
     {
-        Assert.False(JsonPath.TryParse((string)null, out var result));
+        Assert.False(JsonPath.TryParse((string)null!, out var result));
```
meziantou:
```suggestion
        Assert.False(JsonPath.TryParse((string?)null, out var result));
```
If needed update JsonPath.TryParse to allow null values

tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTests.cs
meziantou: Accepted, I changed `_privateField` to `string?` as suggested. This is clearer than using a null-forgiving initializer for this test helper field.

tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTests.cs
meziantou: Accepted, I updated both members to nullable (`_privateFieldIncluded` and `FieldString`) as suggested. This matches their usage in these tests and removes the need for null-forgiving initialization.

tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTests.cs
meziantou: Accepted, I changed `Prop3` to `string?` as suggested. The converter path already handles nullable input, so this better reflects the test model.

tests/Meziantou.Framework.HumanReadableSerializer.Tests/SerializerTestsBase.cs
meziantou: Accepted, I made `Validation.Expected` nullable. I also added an explicit guard in `AssertSerialization(Validation)` so usage remains safe and failures are clear if `Expected` is not provided.

tests/Meziantou.Framework.JsonPath.Tests/JsonPathParseTests.cs
meziantou: Accepted, I updated the test to use `(string?)null` in `TryParse_NullExpression_ReturnsFalse`. No `JsonPath.TryParse` API change was needed because it already accepts nullable input and returns `false` for null.

PR #1333, opened 2026-04-28
Enable nullable reference types in selected test projects

meziantou:
## Why
These test projects had nullable reference types explicitly disabled, which hides potential nullability issues in tests and makes diagnostics less consistent with the rest of the repo.

## What changed
- Removed `<Nullable>disable</Nullable>` from:
  - `Meziantou.Framework.SimpleQueryLanguage.Tests`
  - `Meziantou.Framework.SingleInstance.Tests`
  - `Meziantou.Framework.Slug.Tests`
  - `Meziantou.Framework.SnapshotTesting.Tests`
  - `Meziantou.Framework.StronglyTypedId.Tests`
- Updated `StronglyTypedIdSourceGeneratorTests` to satisfy nullable analysis without changing test behavior by using nullability annotations, `Assert.NotNull(...)`, and null-forgiving where appropriate.

## Notes
- `Meziantou.Framework.StronglyTypedId.GeneratorTests` did not contain `<Nullable>disable</Nullable>` in its csproj, so no csproj change was needed there.
- The `<Nullable>disable</Nullable>` text inside `SnapshotEndToEndTests.cs` is part of test source content and was intentionally left unchanged.

## Validation
- Built all 6 requested test projects.
- Ran tests for all 6 requested test projects.
- Ran required scripts:

### DIFF
--- a/tests/Meziantou.Framework.SimpleQueryLanguage.Tests/ExpressionQueryBuilderTests.cs
+++ b/tests/Meziantou.Framework.SimpleQueryLanguage.Tests/ExpressionQueryBuilderTests.cs
@@ -158,6 +158,6 @@ public sealed class ExpressionQueryBuilderTests
     {
         public int Int32Value { get; set; }
         public long Int64Value { get; set; }
-        public string StringValue { get; set; }
+        public string StringValue { get; set; } = "";
     }
 }
--- a/tests/Meziantou.Framework.SimpleQueryLanguage.Tests/ExpressionQueryBuilderTests.cs
+++ b/tests/Meziantou.Framework.SimpleQueryLanguage.Tests/ExpressionQueryBuilderTests.cs
@@ -158,6 +158,6 @@ public sealed class ExpressionQueryBuilderTests
     {
         public int Int32Value { get; set; }
         public long Int64Value { get; set; }
-        public string StringValue { get; set; } = "";
+        public string? StringValue { get; set; }
     }
 }
--- a/tests/Meziantou.Framework.SimpleQueryLanguage.Tests/QueryBuilderTests.cs
+++ b/tests/Meziantou.Framework.SimpleQueryLanguage.Tests/QueryBuilderTests.cs
@@ -863,6 +863,6 @@ public sealed class QueryBuilderTests
         public DateTime DateTimeValue { get; set; }
         public DateTimeOffset DateTimeOffsetValue { get; set; }
         public DayOfWeek DayOfWeekValue { get; set; }
-        public string StringValue { get; set; } = "";
+        public string? StringValue { get; set; }
     }
 }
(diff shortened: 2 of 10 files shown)


tests/Meziantou.Framework.SimpleQueryLanguage.Tests/ExpressionQueryBuilderTests.cs
```diff
@@ -158,6 +158,6 @@ private sealed class Sample
     {
         public int Int32Value { get; set; }
         public long Int64Value { get; set; }
-        public string StringValue { get; set; }
+        public string StringValue { get; set; } = "";
```
meziantou:
```suggestion
        public string? StringValue { get; set; }
```

tests/Meziantou.Framework.SimpleQueryLanguage.Tests/QueryBuilderTests.cs
```diff
@@ -863,6 +863,6 @@ private sealed class Sample
         public DateTime DateTimeValue { get; set; }
         public DateTimeOffset DateTimeOffsetValue { get; set; }
         public DayOfWeek DayOfWeekValue { get; set; }
-        public string StringValue { get; set; }
+        public string StringValue { get; set; } = "";
```
meziantou:
```suggestion
        public string? StringValue { get; set; }
```

tests/Meziantou.Framework.StronglyTypedId.Tests/StronglyTypedIdSourceGeneratorTests.cs
```diff
@@ -84,7 +84,7 @@ .. dlls.Select(loc => MetadataReference.CreateFromFile(loc)),
             Assert.True(result.Success);
         }
 
-        return (runResult, outputCompilation, result.Success ? outputStream.ToArray() : null, pdbStream.ToArray());
+        return (runResult, outputCompilation, result.Success ? outputStream.ToArray() : null!, pdbStream.ToArray());
```
meziantou: Update the method signature to allow null instead of using `null!`.

tests/Meziantou.Framework.SimpleQueryLanguage.Tests/ExpressionQueryBuilderTests.cs
meziantou: Applied, thanks. I changed `StringValue` to `string?` in this test `Sample` type so nullability matches how test instances are created.

tests/Meziantou.Framework.SimpleQueryLanguage.Tests/QueryBuilderTests.cs
meziantou: Applied here as well. `StringValue` is now `string?` in this test `Sample` type for the same nullability reason.

tests/Meziantou.Framework.StronglyTypedId.Tests/StronglyTypedIdSourceGeneratorTests.cs
meziantou: Good catch. I updated `GenerateFiles` to return `byte[]?` for `Assembly` and replaced `null!` with `null`, keeping behavior unchanged.

PR #1370, opened 2026-05-06
Drop .NET Framework and old netstandard TFMs

### LINKED ISSUE #1282: Proposal: Drop .NET Framework and .NET Standard 2.0 (move to .NET 8+)
I’d like to propose removing .NET Framework and .NET Standard 2.0 support from this repository, and moving to .NET 8+ only.

This is not decided yet. I’m opening this issue to gather feedback from users before making the change.

Note: nuget.org doesn't provide relevant data to know which TFMs are actually used.

# Motivation (maintenance)

Supporting older target frameworks increases ongoing maintenance costs, for example:

- Multi-targeting and conditional compilation complexity
- Larger dependency/build/test matrix
- More time spent on compatibility workarounds and framework-specific issues
- Harder to adopt newer BCL/APIs and runtime improvements cleanly
- Moving to .NET 8+ would simplify maintenance and allow focusing on modern .NET.

More and more projects are using modern .NET, so older TFM may be less important.

# Proposed direction
- Stop targeting netstandard2.0
- Stop targeting any .NET Framework TFMs (net4x, etc.)
- Target net8.0 (and newer) going forward

# Feedback requested
If you use this project, please comment with:

- Which package(s) you use
- Your current target/runtime (e.g., .NET Framework 4.8, netstandard2.0 consumer, etc.)
- Whether you can move to .NET 8+ (and if not, what’s blocking)

If there’s significant dependency on .NET Framework / netstandard2.0, I may adjust the plan/timeline.

meziantou:
## Why

This drops the legacy target frameworks that are no longer meant to be supported in this repo, which simplifies the multi-targeting matrix and lets us remove compatibility code that only existed for those older TFMs.

## What changed

- removed `net462`, `net472`, `net48`, `netstandard2.1`, and general `netstandard2.0` targeting from the affected libraries, samples, and tests
- preserved `netstandard2.0` for Roslyn source generators and the related test wiring that still depends on it
- deleted obsolete `#if NETFRAMEWORK` and old `NETSTANDARD*` branches that were only needed for the dropped TFMs
- updated shared validation and build configuration so the new target-framework expectations are enforced consistently
- merged the latest `origin/main` and resolved the `SnapshotTesting.Common` project deletion in favor of the upstream removal

## Notes for reviewers

The one intentional exception is Roslyn source generators: they still keep `netstandard2.0` because some test and analyzer flows explicitly depend on that target.

This also includes the validator fix needed so those generator references do not incorrectly fail test-project configuration checks.

Fix https://github.com/meziantou/Meziantou.Framework/issues/1282


tests/Meziantou.Framework.InlineSnapshotTesting.Tests/SnapshotSerializerTests.cs
```diff
@@ -3,16 +3,12 @@
...
-using System.Net.Http;
-#endif
-
 namespace Meziantou.Framework.InlineSnapshotTesting.Tests;
 public sealed class SnapshotSerializerTests
 {
     private static readonly InlineSnapshotSettings Settings = InlineSnapshotSettings.Default with
     {
-        ValidateLineNumberUsingPdbInfoWhenAvailable = false, // Fail on net472
+        ValidateLineNumberUsingPdbInfoWhenAvailable = false,
```
meziantou: Can be removed as there is no net472 anymore

src/Meziantou.Framework.WPF/Meziantou.Framework.WPF.csproj
```diff
@@ -1,19 +1,15 @@
...
+    <TargetFrameworks>$(LatestTargetFrameworksWindows)</TargetFrameworks>
     <IsTrimmable>false</IsTrimmable>
     <UseWPF>true</UseWPF>
-    <Version>1.5.25</Version>
+    <Version>2.0.0</Version>
     <Description>Collection of Commands, Markup extensions, DependencyProperty, and collections for WPF</Description>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="System.Collections.Immutable" Version="10.0.7" />
```
meziantou: Is it needed?

PR #1405, opened 2026-05-09
Add Meziantou.Xunit conditional test attributes

meziantou:
## Why
Test-only condition helpers were living in `tests/TestUtilities`, making them hard to reuse and evolve. This change introduces a dedicated xUnit v3 helper package with a single shared condition engine for run/skip behavior.

## What changed
- Added new package project: `src/Meziantou.Xunit`.
- Moved conditional test helpers out of `tests/TestUtilities` into `Meziantou.Xunit`.
- Added unified `RunIf` and `SkipIf` behavior on top of shared evaluation logic.
- Added condition enums and helpers:
  - `TestOperatingSystems`
  - `TestGlobalizationMode`
  - `ContinuousIntegrationEnvironment`
  - `WindowsGroups` (`Any`, `User`, `Administrator`)
- Merged the administrator-only attribute into `RunIf`/`SkipIf` conditions:
  - `RunIf(WindowsGroups.Administrator)`
- Replaced old test usages (`Fact*`, `SkipOnGitHubActions`, `RunIfWindowsAdministrator`) with the new API.
- Kept NuGet helpers in `tests/TestUtilities` as requested.
- Updated solution/readme/generated project files via repository scripts.

## Validation
- Ran required scripts:
- Built solution: `dotnet build Meziantou.Framework.slnx --no-restore`
- Ran impacted tests with `DiffEngine_Disabled=true`, including:
  - `tests/Meziantou.Framework.Tests`
  - `tests/Meziantou.Framework.TypeConverter.Tests`
  - `tests/Meziantou.Framework.Http.Caching.Tests`
  - `tests/Meziantou.Framework.Win32.Lsa.Tests`
  - `tests/Meziantou.Framework.Win32.ChangeJournal.Tests`


src/Meziantou.Xunit/RunIfAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+        : base(windowsGroup)
+    {
+    }
+
+    public RunIfAttribute(TestOperatingSystems operatingSystem, TestGlobalizationMode globalizationMode)
+        : base(operatingSystem, globalizationMode)
+    {
+    }
+
+    public static bool IsGlobalizationInvariant() => TestEnvironment.IsGlobalizationInvariant();
```
meziantou: This should not be exposed directly. If another project needs it, share the code by MSBuild.

src/Meziantou.Xunit/ContinuousIntegrationEnvironment.cs
```diff
@@ -0,0 +1,8 @@
+namespace Meziantou.Xunit;
+
+[Flags]
+public enum ContinuousIntegrationEnvironment
```
meziantou:
```suggestion
public enum ContinuousIntegrationEnvironments
```

src/Meziantou.Xunit/RunIfAttribute.cs
meziantou: Good catch. I removed `RunIfAttribute.IsGlobalizationInvariant()` so this helper is no longer exposed on the attribute API. If this logic is needed elsewhere, we can share it through MSBuild instead.

src/Meziantou.Xunit/ContinuousIntegrationEnvironment.cs
meziantou: Good suggestion, thanks. I renamed the flags enum to `ContinuousIntegrationEnvironments` and updated all references.

PR #1420, opened 2026-05-10
Add OpenTelemetry request and tail filtering hooks

meziantou:
## Why
OpenTelemetry traces can arrive out of order (for example, child spans before the root span), which makes simple head filtering insufficient for some sampling scenarios. This change adds first-class filtering hooks so applications can filter logs/metrics/traces and perform trace tail filtering with bounded in-memory buffering.

## What changed
- Added new public collector API:
  - `OpenTelemetryReceiverOptions`
  - `OpenTelemetryRequestFilter<TRequest>` for logs/traces/metrics request filtering
  - `OpenTelemetryTailSamplingOptions`, `OpenTelemetryTailBufferOverflowPolicy`, `OpenTelemetryTraceTailFilter`, and `OpenTelemetryTailTraceContext` for tail filtering
- Replaced internal options plumbing with public receiver options and added optional configuration callbacks to `AddOpenTelemetryReceiver` overloads.
- Introduced a shared request pipeline (`OpenTelemetryRequestPipeline`) used by both HTTP and gRPC entry points.
- Added tail sampling implementation (`OpenTelemetryTraceTailSampler`) that:
  - buffers spans per trace id
  - completes evaluation when root span arrives or when `MaxTraceDuration` is reached
  - enforces `MaxBufferedSpansPerTrace` and `MaxBufferedSpans`
  - applies configurable overflow behavior (`DropWholeTrace`, `DropOldestSpans`, `DropNewestSpans`)
- Updated OTLP trace proto to include `parent_span_id` to support root-span detection.
- Added in-memory receiver overloads to configure `OpenTelemetryReceiverOptions` directly.
- Updated package README with request filtering and tail filtering examples.

## Notes for reviewers
- Filtered-out payloads still return normal OTLP success responses so exporters do not retry dropped telemetry.
- Tail filtering is trace-specific; logs and metrics continue using request-level filters.

## Validation
- `DiffEngine_Disabled=true dotnet test tests/Meziantou.Framework.OpenTelemetry.Tests/Meziantou.Framework.OpenTelemetry.Tests.csproj`


src/Meziantou.Framework.OpenTelemetryCollector/OpenTelemetryTraceTailSampler.cs
```diff
@@ -0,0 +1,312 @@
...
+
+internal sealed class OpenTelemetryTraceTailSampler(IOptions<OpenTelemetryReceiverOptions> optionsAccessor)
+{
+#if NET9_0_OR_GREATER
+    private readonly System.Threading.Lock _gate = new();
+#else
+#pragma warning disable MA0158 // System.Threading.Lock is not available on net8
+    private readonly object _gate = new();
+#pragma warning restore MA0158
+#endif
```
meziantou: No need for the `#if`. There are polyfills for older TFM

meziantou: Good point. Removed the conditional compilation and now always use System.Threading.Lock.

src/Meziantou.Framework.OpenTelemetryCollector/OpenTelemetryTailSampling.cs
```diff
@@ -0,0 +1,14 @@
+namespace Meziantou.Framework.OpenTelemetryCollector;
+
+public sealed class OpenTelemetryTailSampling : OpenTelemetrySampler
```
meziantou:
```suggestion
public sealed class OpenTelemetryTailSampler : OpenTelemetrySampler
```

src/Meziantou.Framework.OpenTelemetryCollector/OpenTelemetryTraceTailSampler.cs
```diff
@@ -0,0 +1,308 @@
+using OpenTelemetry.Proto.Common.V1;
+using OpenTelemetry.Proto.Collector.Trace.V1;
+using OpenTelemetry.Proto.Resource.V1;
+using OpenTelemetry.Proto.Trace.V1;
+
+namespace Meziantou.Framework.OpenTelemetryCollector;
+
+internal sealed class OpenTelemetryTraceTailSampler
```
meziantou:
```suggestion
internal sealed class OpenTelemetryTraceTailSamplerHandler
```

src/Meziantou.Framework.OpenTelemetryCollector/OpenTelemetryTraceTailSampler.cs
meziantou: Accepted. I removed the conditional compilation and we now always use `System.Threading.Lock`, relying on the polyfills for older TFMs.

meziantou: Accepted. This is still the current implementation in the branch, so no additional follow-up change was needed.

src/Meziantou.Framework.OpenTelemetryCollector/OpenTelemetryTailSampling.cs
meziantou: Accepted. I renamed the public type to `OpenTelemetryTailSampler` and updated all related usages in code, tests, and documentation.

src/Meziantou.Framework.OpenTelemetryCollector/OpenTelemetryTraceTailSampler.cs
meziantou: Accepted. I renamed the internal class to `OpenTelemetryTraceTailSamplerHandler` and updated the dependent DI and pipeline references.

PR #1421, opened 2026-05-10
Rename and expand FastEnum generator

meziantou:
This change renames the enum source generator package to match its broader scope and extends it beyond `ToStringFast` so consumers can use generated enum helpers consistently from one package.

## What changed
- Renamed `Meziantou.Framework.FastEnumToStringGenerator` to `Meziantou.Framework.FastEnumGenerator` across source, tests, slnx entries, and package references (`README`, `renovate.json5`, project paths, and readme links).
- Replaced the old generator with `FastEnumSourceGenerator`, which now emits:
  - `ToStringFast(this TEnum)` and `ToStringFast(this TEnum, bool useMetadata)`
  - `HasFlag(this TEnum instance, TEnum flag)`
  - `GetName(this TEnum instance)`
  - C# 14 extension members (when supported): `Parse`, `TryParse`, `IsDefined`, `GetNames`, `GetValues`
- Added metadata-aware behavior (`DisplayAttribute` and `EnumMemberAttribute`) and flags-aware parsing/formatting paths.
- Kept compatibility with the legacy `[FastEnumToStringAttribute]` while adding `[FastEnumAttribute]`.
- Expanded generator tests to cover rename impact, metadata/flags behavior, and C# 14 gating behavior.

## Non-obvious decisions
- Static enum APIs are emitted only when extension members are available, based on target language support.
- Generated types are split per enum (`FastEnumExtensions_<index>`) to avoid method-signature collisions when several enums share the same namespace/type container while keeping extension-method discovery intact.

## Validation
- `dotnet build slnx/Meziantou.Framework.FastEnumGenerator.slnx`
- `DiffEngine_Disabled=true dotnet test tests/Meziantou.Framework.FastEnumGenerator.Tests/Meziantou.Framework.FastEnumGenerator.Tests.csproj`


src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
```diff
@@ -0,0 +1,834 @@
...
+    }
+
+    private static bool SupportsExtensionMembers(Compilation compilation)
+    {
+        if (compilation is not CSharpCompilation csharpCompilation)
+            return false;
+
+        var parseOptions = csharpCompilation.SyntaxTrees.FirstOrDefault()?.Options as CSharpParseOptions ?? CSharpParseOptions.Default.WithLanguageVersion(LanguageVersion.Preview);
+        var tree = CSharpSyntaxTree.ParseText("static class __Probe { extension(int value) { public static int Parse(string value, bool ignoreCase) => 0; } }", parseOptions);
+        return tree.GetDiagnostics().All(static diagnostic => diagnostic.Severity != DiagnosticSeverity.Error);
```
meziantou: Use the FirstSyntaxTree options. If no SyntaxTree, return false. Anyway there will be no code to generate.

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
```diff
@@ -0,0 +1,834 @@
...
+        var generationInput = enumToProcess.Combine(supportsExtensionMembers);
+
+        context.RegisterSourceOutput(generationInput, static (spc, source) => GenerateCode(spc, source.Left, source.Right));
+    }
+
+    private static EnumToProcess? GetSemanticTargetForGeneration(GeneratorSyntaxContext ctx, CancellationToken cancellationToken)
+    {
+        var compilation = ctx.SemanticModel.Compilation;
+        var fastEnumAttributeSymbol = compilation.GetTypeByMetadataName("Meziantou.Framework.Annotations.FastEnumAttribute");
+        var legacyFastEnumAttributeSymbol = compilation.GetTypeByMetadataName("Meziantou.Framework.Annotations.FastEnumToStringAttribute");
```
meziantou: Remove legacy support

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
```diff
@@ -0,0 +1,834 @@
...
+        sb.AppendLine("        if (value.Length != token.Length)");
+        sb.AppendLine("            return false;");
+        sb.AppendLine();
+        sb.AppendLine("        if (!ignoreCase)");
+        sb.AppendLine("        {");
+        sb.AppendLine("            for (var i = 0; i < token.Length; i++)");
+        sb.AppendLine("            {");
+        sb.AppendLine("                if (value[i] != token[i])");
+        sb.AppendLine("                    return false;");
+        sb.AppendLine("            }");
```
meziantou: You can use Equals method with StringComparison.Ordinal / OrdinalIgnoreCase instead of char by char comparison

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
meziantou: Done. I updated `SupportsExtensionMembers` to use the first syntax tree parse options. If there is no syntax tree (or no C# parse options), it now returns `false` and skips extension-member generation.

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
meziantou: Done. I removed legacy `FastEnumToStringAttribute` support from the generator and analyzer paths, and removed the legacy compatibility test. Only `FastEnumAttribute` is supported now.

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
```diff
@@ -0,0 +1,834 @@
...
+            if (enumType.TypeKind != TypeKind.Enum)
+                continue;
+
+            var members = GetMembers(enumType);
+            if (members.Length == 0)
+                continue;
+
+            var isPublic = GetIsPublic(attr, enumType);
+            var extensionNamespace = GetExtensionNamespace(attr);
+            var isFlags = enumType.GetAttributes().Any(static attribute => string.Equals(attribute.AttributeClass?.ToDisplayString(), "System.FlagsAttribute", StringComparison.Ordinal));
```
meziantou: Compare by ISymbol, not display string.

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+        defaultSeverity: DiagnosticSeverity.Error,
+        isEnabledByDefault: true);
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(InvalidEnumType);
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+        context.EnableConcurrentExecution();
+        context.RegisterSyntaxNodeAction(static context => AnalyzeAttribute(context), SyntaxKind.Attribute);
```
meziantou: Can it use IOperation with IAttributeOperation instead of Syntax?
Or maybe use RegisterSymbol (assembly, type) to find attributes?

Also, use CompilationStart to find the attribute, then register actions.

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
```diff
@@ -0,0 +1,820 @@
...
+
+        var firstSyntaxTree = csharpCompilation.SyntaxTrees.FirstOrDefault();
+        if (firstSyntaxTree is null)
+            return false;
+
+        if (firstSyntaxTree.Options is not CSharpParseOptions parseOptions)
+            return false;
+
+        var tree = CSharpSyntaxTree.ParseText("static class __Probe { extension(int value) { public static int Parse(string value, bool ignoreCase) => 0; } }", parseOptions);
+        return tree.GetDiagnostics().All(static diagnostic => diagnostic.Severity != DiagnosticSeverity.Error);
```
meziantou: Don't use the probe. Check the parseOption language version directly

.github/workflows/ci.yml
```diff
@@ -245,6 +245,10 @@ jobs:
             !**/*.coverage
             **/*.binlog
             ${{ env.TestResultsDirectory }}/**/*
+            tests/Meziantou.AspNetCore.ServiceDefaults.AutoRegister.GeneratorTests/obj/**/GeneratedFiles/**/*
+            tests/Meziantou.Framework.FastEnumGenerator.GeneratorTests/obj/**/GeneratedFiles/**/*
+            tests/Meziantou.Framework.ResxSourceGenerator.GeneratorTests/obj/**/GeneratedFiles/**/*
+            tests/Meziantou.Framework.StronglyTypedId.GeneratorTests/obj/**/GeneratedFiles/**/*
```
meziantou:
```suggestion
            **/obj/**/GeneratedFiles/**/*
```

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
meziantou: Good call, I updated this. The generated token comparison now uses `MemoryExtensions.Equals` with `StringComparison.Ordinal` / `OrdinalIgnoreCase` instead of the manual char-by-char loop.

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
meziantou: Agreed, I changed this to symbol comparison. The generator now resolves `System.FlagsAttribute` from the compilation and uses `SymbolEqualityComparer.Default` instead of comparing display strings.

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
meziantou: Partially agreed and updated. I moved the analyzer to `RegisterCompilationStartAction` and then do symbol-based analysis from compilation data; I did not use `IOperation` because this case is assembly-level attributes and the symbol path is simpler and cheaper.

src/Meziantou.Framework.FastEnumGenerator/FastEnumSourceGenerator.cs
meziantou: Agreed, I removed the probe-based check. `SupportsExtensionMembers` now reads the first syntax tree parse options directly and compares `LanguageVersion`, returning false when there is no syntax tree.

.github/workflows/ci.yml
meziantou: I am declining this one for now and keeping explicit project-scoped paths. The broad `**/obj/**/GeneratedFiles/**/*` pattern would also pick up generated files outside the intended generator test projects and can significantly increase artifact volume.

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
```diff
@@ -0,0 +1,65 @@
...
+        context.RegisterCompilationStartAction(InitializeCore);
+    }
+
+    private static void InitializeCore(CompilationStartAnalysisContext context)
+    {
+        var fastEnumAttribute = context.Compilation.GetTypeByMetadataName("Meziantou.Framework.Annotations.FastEnumAttribute");
+        if (fastEnumAttribute is null)
+            return;
+
+        context.RegisterCompilationEndAction(context => AnalyzeCompilation(context, fastEnumAttribute));
```
meziantou: Never use RegisterCompilationEnd. Use RegisterOperation(OperationKind.Attribute).

Also, inline the InitializeCore directly in the RegisterCompilationStartAction method

tests/Meziantou.Framework.FastEnumGenerator.GeneratorTests/FastEnumGeneratedBehaviorTests.cs
```diff
@@ -0,0 +1,45 @@
+using Meziantou.Framework.FastEnumGenerator.GeneratorTests.Generated;
+using Xunit;
+
+[assembly: Meziantou.Framework.Annotations.FastEnumAttribute(typeof(Meziantou.Framework.FastEnumGenerator.GeneratorTests.Color), ExtensionMethodNamespace = "Meziantou.Framework.FastEnumGenerator.GeneratorTests.Generated")]
+[assembly: Meziantou.Framework.Annotations.FastEnumAttribute(typeof(Meziantou.Framework.FastEnumGenerator.GeneratorTests.Permission), ExtensionMethodNamespace = "Meziantou.Framework.FastEnumGenerator.GeneratorTests.Generated")]
+
+namespace Meziantou.Framework.FastEnumGenerator.GeneratorTests;
+
+public sealed class FastEnumGeneratedBehaviorTests
```
meziantou, line 22: I think you can add more tests. For instance,
- An enum with multiple names for the same value
- Flags with defined combination (None, A = 1, B = 2, AandB = 3, C = 4)
- Be sure to test all methods

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
meziantou: Accepted. I replaced the compilation-end path with `RegisterOperationAction(..., OperationKind.Attribute)` and inlined the compilation-start setup directly in `Initialize`; I also updated `Microsoft.CodeAnalysis` to 5.3.0 because the 4.2.0 API surface here does not expose attribute operations.

tests/Meziantou.Framework.FastEnumGenerator.GeneratorTests/FastEnumGeneratedBehaviorTests.cs
meziantou: Accepted. I added behavior tests for both requested scenarios (enum aliases and flags with a defined `AandB = 3` combination), and expanded coverage so all generated methods and Parse/TryParse overloads are exercised.

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private static void AnalyzeAttribute(OperationAnalysisContext context, IObjectCreationOperation attribute)
+    {
+        if (attribute.Arguments.Length != 1)
+            return;
+
+        var location = attribute.Syntax.GetLocation();
+        var argument = attribute.Arguments[0].Value;
+        while (argument is IConversionOperation conversionOperation)
```
meziantou: Why would there be a conversion for the type?

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+            argument = conversionOperation.Operand;
+        }
+
+        if (argument.ConstantValue is { HasValue: true, Value: null })
+        {
+            context.ReportDiagnostic(Diagnostic.Create(InvalidEnumType, location, "(null)"));
+            return;
+        }
+
+        if (argument is not ITypeOfOperation typeOfOperation || typeOfOperation.TypeOperand.TypeKind == TypeKind.Enum)
```
meziantou:
```suggestion
        if (argument is ITypeOfOperation { TypeOperand.TypeKind is TypeKind.Enum } typeOfOperation)
```

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+                return;
+
+            context.RegisterOperationAction(context => AnalyzeAttributeOperation(context, fastEnumAttribute), OperationKind.Attribute);
+        });
+    }
+
+    private static void AnalyzeAttributeOperation(OperationAnalysisContext context, INamedTypeSymbol fastEnumAttribute)
+    {
+        if (context.Operation is not IAttributeOperation attributeOperation)
+            return;
```
meziantou: No need for safe-cast
```suggestion
        var attributeOperation = (IAttributeOperation)context.Operation;
```

src/Meziantou.Framework.FastEnumGenerator/Meziantou.Framework.FastEnumGenerator.csproj
```diff
@@ -2,19 +2,19 @@
...
 
   <ItemGroup>
-    <None Include="Meziantou.Framework.FastEnumToStringGenerator.targets" Pack="true" PackagePath="build/" />
+    <None Include="Meziantou.Framework.FastEnumGenerator.targets" Pack="true" PackagePath="build/" />
     <None Include="$(OutputPath)/../$(Configuration.ToLowerInvariant())_netstandard2.0/$(AssemblyName).dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.2.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="5.3.0" PrivateAssets="all" />
```
meziantou: Does it need 5.3.0 or is 5.0.0 enough?

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
meziantou: Accepted. You are right that this path does not need conversion unwrapping, so I removed that loop and now use the argument value directly.

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
meziantou: Partially accepted. I switched this to property-pattern matching for `ITypeOfOperation`, using the inverted condition so we keep reporting only non-enum `typeof(...)` operands.

src/Meziantou.Framework.FastEnumGenerator/FastEnumAnalyzer.cs
meziantou: Accepted. I replaced the safe cast with a direct cast to `IAttributeOperation`, since this callback is already registered for `OperationKind.Attribute`.

src/Meziantou.Framework.FastEnumGenerator/Meziantou.Framework.FastEnumGenerator.csproj
meziantou: Accepted. 5.0.0 is enough for this analyzer path, so I downgraded from 5.3.0 to 5.0.0 and revalidated with build and the FastEnum test suites.

PR #1433, opened 2026-05-11
Optimize hot-path parsing and protocol components

meziantou:
This PR implements a set of planned performance optimizations across parser and protocol hot paths to reduce allocations and per-call overhead while keeping behavior unchanged.

## What changed

- `CsvReader` now uses buffered async reads with internal lookahead instead of single-char async reads.
- `BitBuffer` now uses a growable `byte[]` with a byte-aligned append fast path.
- `HtmlSanitizer` default allow/block sets are precomputed as `FrozenSet`; `UrlSanitizer.IsSafeSrcset` now uses span-based parsing.
- `CharacterSetSegment` now uses precomputed `SearchValues<char>` matching.
- `RegexScanner` now caches a compiled `Regex` instance and resolved group indices.
- `TdsPacketReader` and `DnsWireReader` now use pooled buffers to cut packet/name decoding allocations.
- `NtpPacket` switched manual endian bit-shifts to `BinaryPrimitives` read/write helpers.
- `SimpleQuery` lexer now uses `SearchValues<char>` token boundary detection and span-based keyword matching.
- `FileLogger` now caches the short category name and builds each line through one builder path.
- `ObjectMethodExecutor` now exposes `GetOrCreate(MethodInfo)` backed by a weak cache, with test coverage.

## Notes for review

- This is a behavior-preserving refactor set focused on throughput and GC pressure.
- `RegexScanner` now fails fast if the configured pattern does not define the required `name` group.

### DIFF
--- a/src/Meziantou.Framework.DependencyScanning/Scanners/RegexScanner.cs
+++ b/src/Meziantou.Framework.DependencyScanning/Scanners/RegexScanner.cs
@@ -25,6 +25,10 @@ public sealed class RegexScanner : DependencyScanner
     private const string VersionGroupName = "version";
 
     private bool _frozen;
+    private string? _regexPattern;
+    private Regex? _regex;
+    private int _nameGroupNumber = -1;
+    private int _versionGroupNumber = -1;
 
     protected internal override IReadOnlyCollection<DependencyType> SupportedDependencyTypes
     {
@@ -37,7 +41,20 @@ public sealed class RegexScanner : DependencyScanner
     }
 
     /// <summary>Gets or sets the regular expression pattern to match dependencies. The pattern must include named groups 'name' and optionally 'version'.</summary>
-    public string? RegexPattern { get; set; }
+    public string? RegexPattern
+    {
+        get => _regexPattern;
+        set
+        {
+            if (_frozen)
+                throw new InvalidOperationException("The scanner is already used and cannot be modified.");
+
+            _regexPattern = value;
+            _regex = null;
+            _nameGroupNumber = -1;
+            _versionGroupNumber = -1;
+        }
+    }
 
     /// <summary>Gets or sets the type of dependency to report when a match is found.</summary>
     public DependencyType DependencyType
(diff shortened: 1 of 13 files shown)


src/Meziantou.Framework.DependencyScanning/Scanners/RegexScanner.cs
```diff
@@ -57,35 +74,49 @@ public DependencyType DependencyType
...
     public override async ValueTask ScanAsync(ScanFileContext context)
     {
-        if (RegexPattern is null)
+        var regexPattern = RegexPattern;
+        if (regexPattern is null)
             return;
 
         _frozen = true;
+        var regex = _regex;
+        if (regex is null)
```
meziantou, line 83: The scanner can be used in parallel. Maybe a lock would be better here?
Or maybe it should replace the pattern property with a Regex property, so the caller can provide the regex with the expected options, and even use generated regex.

src/Meziantou.Framework.DnsClient/Protocol/DnsWireReader.cs
```diff
@@ -94,53 +95,68 @@ private static void ReadDomainNameCore(StringBuilder sb, ReadOnlySpan<byte> mess
         var jumped = false;
         var originalPosition = -1;
         var pointerCount = 0;
+        var charBuffer = ArrayPool<char>.Shared.Rent(63);
```
meziantou, line 98: Can it use stackalloc with a Span<char>?

src/Meziantou.Framework.ObjectMethodExecutor/ObjectMethodExecutor.cs
```diff
@@ -47,6 +47,7 @@ public sealed class ObjectMethodExecutor
             typeof(Action<object, Action>), // onCompletedMethod
             typeof(Action<object, Action>), // unsafeOnCompletedMethod
         ])!;
+    private static readonly ConditionalWeakTable<MethodInfo, ObjectMethodExecutor> MethodExecutorCache = new();
```
meziantou, line 50: Revert this file. The caller should cache the result manually. It's not this class responsibility

src/Meziantou.Framework.Globbing/Internals/Segments/CharacterSetSegment.cs
```diff
@@ -1,16 +1,19 @@
+using System.Buffers;
```
meziantou, line 1: revert change in this file. I will do better changes in the future

PR #1436, opened 2026-05-11
Refactor BitBuffer to use array instead of List

meziantou:


src/Meziantou.Framework.QRCode/Internal/BitBuffer.cs
```diff
@@ -2,34 +2,68 @@ namespace Meziantou.Framework.Internal;
 
 internal sealed class BitBuffer
 {
-    private readonly List<byte> _bytes = [];
+    private byte[] _bytes = [];
```
meziantou: I think it can be byte[]? _bytes = null to avoid one useless assignation. EnsureCapacity should allocate the initial array

meziantou: Done in f2a70afe: `_bytes` is now nullable and initialized in `EnsureCapacity` (first allocation is performed there).

PR #1526, opened 2026-05-19
Simplify HTTP Basic credential validation API

meziantou:
HTTP Basic authentication currently splits credential validation and principal creation across two delegates. This makes configuration more complex and causes redundant user lookups in the Identity integration.

This change makes `ValidateCredentials` the single extension point: it now returns `ValueTask<ClaimsPrincipal?>`, where `null` means authentication failure. The handler now authenticates through a single delegate call, and the Identity integration validates credentials and creates the principal in one flow.

Additional updates:
- remove `CreatePrincipal` from `HttpBasicAuthenticationOptions`
- delete `HttpBasicPrincipalFactory`
- update tests and README examples to the new contract
- regenerate `ref/Meziantou.AspNetCore.Authentication.HttpBasic.cs`

Breaking change: custom validators must now return a `ClaimsPrincipal?` instead of `bool`.


src/Meziantou.AspNetCore.Authentication.HttpBasic/HttpBasicAuthenticationOptions.cs
```diff
@@ -27,25 +27,8 @@ public int MaxCredentialLength
...
-        var claims = new[]
-        {
-            new Claim(ClaimTypes.Name, username),
-            new Claim(ClaimTypes.NameIdentifier, username),
-        };
-
-        var identity = new ClaimsIdentity(claims, authenticationType: authenticationScheme);
-        return new ClaimsPrincipal(identity);
-    }
+    public HttpBasicCredentialValidator ValidateCredentials { get; set; } = static (_, _, _) => ValueTask.FromResult<ClaimsPrincipal?>(null);
```
meziantou:
```suggestion
    public HttpBasicCredentialValidator ValidateCredentials { get; set; } = (_, _, _) => ValueTask.FromResult<ClaimsPrincipal?>(null);
```

src/Meziantou.AspNetCore.Authentication.HttpBasic/readme.md
```diff
@@ -19,10 +20,21 @@ builder.Services
     {
         options.Realm = "My application";
         options.MaxCredentialLength = 4096;
-        options.ValidateCredentials = static (context, username, password) =>
+        options.ValidateCredentials = static (_, username, password) =>
```
meziantou:
```suggestion
        options.ValidateCredentials = (context, username, password) =>
```

PR #1550, opened 2026-05-23
Remove LibGit2Sharp from Git submodule dependency scanner

meziantou:
## Why
`GitSubmoduleDependencyScanner` depended on LibGit2Sharp only to read submodule metadata and pinned SHAs. Replacing it reduces package dependencies while keeping the same feature.

## What changed
- Replaced LibGit2Sharp usage in `GitSubmoduleDependencyScanner` with a pure managed implementation.
- Added `.gitmodules` parsing to read submodule `path` and `url` entries.
- Added `.git/index` parsing to extract gitlink entries (`mode 160000`) and map submodule path to pinned SHA.
- Added support for `.git` indirection files (`gitdir: ...`) when locating the index.
- Preserved resilient behavior by skipping reporting when repository/index data cannot be read or is unsupported.
- Removed the `LibGit2Sharp` package reference from `Meziantou.Framework.DependencyScanning.csproj`.
- Updated scanner tests to stop using LibGit2Sharp and resolve the expected SHA via `git rev-parse HEAD`.
- Made `GitSubmodulesFromDependencies` run cross-platform (`[Fact]` instead of Windows-only).

## Validation
- Ran required maintenance scripts:
- Built dependency scanning project with warnings as errors.
- Ran dependency scanning test suite across target frameworks with warnings as errors.

### DIFF
--- a/src/Meziantou.Framework.DependencyScanning/Scanners/GitSubmoduleDependencyScanner.cs
+++ b/src/Meziantou.Framework.DependencyScanning/Scanners/GitSubmoduleDependencyScanner.cs
@@ -136,12 +136,59 @@ public sealed class GitSubmoduleDependencyScanner : DependencyScanner
 
     private static string Unquote(string value)
     {
+        string unquotedValue;
         if (value.Length >= 2 && value[0] is '"' && value[^1] is '"')
         {
-            return value[1..^1];
+            unquotedValue = value[1..^1];
+        }
+        else
+        {
+            unquotedValue = value;
         }
 
-        return value;
+        return UnescapeGitConfigValue(unquotedValue);
+    }
+
+    private static string UnescapeGitConfigValue(string value)
+    {
+        if (value.AsSpan().IndexOf('\\') < 0)
+            return value;
+
+        var builder = new System.Text.StringBuilder(value.Length);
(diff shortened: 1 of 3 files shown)


src/Meziantou.Framework.DependencyScanning/Scanners/GitSubmoduleDependencyScanner.cs
```diff
@@ -15,20 +19,312 @@ protected override bool ShouldScanFileCore(CandidateFileContext context)
...
+        catch (UnauthorizedAccessException)
+        {
+            result = [];
+            return false;
+        }
+    }
+
+    private static bool TryReadExactly(Stream stream, Span<byte> buffer)
+    {
+        var totalRead = 0;
```
meziantou: Use Stream.ReadAtLeast method for implementation.

meziantou: Accepted, this is a good simplification. I updated `TryReadExactly` to use `Stream.ReadAtLeast(..., throwOnEndOfStream: false)` and kept the same behavior by returning `false` when fewer bytes are read.

PR #1710, opened 2026-06-28
Add disk caching for DNS block lists

meziantou:
## Summary
- Add a configurable disk cache for downloaded block lists, with a sensible default under local application data
- Load cached filter lists at startup and refresh the cache after successful downloads
- Surface the cache path in diagnostics and document the new setting
- Extend unit coverage for defaults, diagnostics output, and cache behavior

## Testing
- Unit tests updated for option defaults, diagnostics rendering, and filter engine cache loading/writing
- Not run (not requested)

### DIFF
--- a/tests/Meziantou.Framework.DnsProxy.Tests/FilterEngineProviderTests.cs
+++ b/tests/Meziantou.Framework.DnsProxy.Tests/FilterEngineProviderTests.cs
@@ -2,6 +2,7 @@ using System.Diagnostics;
 using System.Net;
 using Meziantou.DnsProxy;
 using Meziantou.DnsProxy.Filtering;
+using Meziantou.Framework;
 using Meziantou.Framework.DnsFilter;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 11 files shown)


tests/Meziantou.Framework.DnsProxy.Tests/FilterEngineProviderTests.cs
```diff
@@ -173,6 +254,14 @@ private static ServiceProvider CreateServiceProvider(Func<HttpRequestMessage, Ht
         return services.BuildServiceProvider();
     }
 
+    private static string CreateTemporaryDirectory()
+    {
+        var path = Path.Combine(Path.GetTempPath(), "meziantou-dnsproxy-tests-" + Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(path);
+
+        return path;
+    }
```
meziantou: Use TemporaryDirectory from Meziantou.Framework.TemporaryDirectory project directly

PR #1716, opened 2026-06-29
Fix MFAS assertion warnings failing CI build

meziantou:
## Why
The build on `main` started failing because new assertion analyzer rules are treated as errors in CI. Several tests were still using generic count and null-check assertion patterns that now trigger `MFAS0005` and `MFAS0007`.

## What changed
This updates the affected tests to use the analyzer-preferred assertion APIs:
- Replaced length/count comparisons with specialized count assertions (`Assert.HasCount`, `Assert.HasCountGreaterThanOrEqual`)
- Replaced `Assert.True(value is not null)`-style checks with `Assert.NotNull` patterns
- Kept test intent and behavior unchanged while making the assertions analyzer-compliant

## Notes for reviewers
`Meziantou.Framework.Win32.ProjectedFileSystem.Tests` could not be executed in this local environment because the x64 .NET test host is unavailable on this machine. The rest of the touched test projects were run successfully.

### DIFF
--- a/tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableArrayTests.cs
+++ b/tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableArrayTests.cs
@@ -93,7 +93,10 @@ public sealed class ImmutableEquatableArrayTests
         Assert.False(array.Equals(null));
         Assert.False(array == null);
         Assert.False(null == array);
-        Assert.NotNull(array);
+#pragma warning disable MFAS0007 // Preserve == and != operator validation
+        Assert.True(array != null);
+        Assert.True(null != array);
+#pragma warning restore MFAS0007
     }
 
     [Fact]
--- a/tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableArrayTests.cs
+++ b/tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableArrayTests.cs
@@ -93,8 +93,7 @@ public sealed class ImmutableEquatableArrayTests
         Assert.False(array.Equals(null));
         Assert.False(array == null);
         Assert.False(null == array);
-        Assert.True(array != null);
-        Assert.True(null != array);
+        Assert.NotNull(array);
     }
 
     [Fact]
--- a/tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableDictionaryTests.cs
+++ b/tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableDictionaryTests.cs
@@ -181,8 +181,7 @@ public sealed class ImmutableEquatableDictionaryTests
         Assert.False(dict.Equals(null));
         Assert.False(dict == null);
         Assert.False(null == dict);
-        Assert.True(dict != null);
-        Assert.True(null != dict);
+        Assert.NotNull(dict);
         Assert.False(dict.Equals((object?)null));
     }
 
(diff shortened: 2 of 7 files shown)


tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableArrayTests.cs
```diff
@@ -93,8 +93,7 @@ public void Equals_WithNull_ShouldReturnFalse()
         Assert.False(array.Equals(null));
         Assert.False(array == null);
         Assert.False(null == array);
-        Assert.True(array != null);
-        Assert.True(null != array);
+        Assert.NotNull(array);
```
meziantou: Preserve old behavior. It really needs to validate the `==` and `!=` operators. Disable warning explicitly.

tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableDictionaryTests.cs
```diff
@@ -181,8 +181,7 @@ public void Equals_WithNull_ShouldReturnFalse()
         Assert.False(dict.Equals(null));
         Assert.False(dict == null);
         Assert.False(null == dict);
-        Assert.True(dict != null);
-        Assert.True(null != dict);
+        Assert.NotNull(dict);
```
meziantou: Preserve old behavior. It really needs to validate the `==` and `!=` operators. Disable warning explicitly.

tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableSetTests.cs
```diff
@@ -89,8 +89,7 @@ public void Equals_WithNull_ShouldReturnFalse()
         Assert.False(set.Equals(null));
         Assert.False(set == null);
         Assert.False(null == set);
-        Assert.True(set != null);
-        Assert.True(null != set);
+        Assert.NotNull(set);
```
meziantou: Preserve old behavior. It really needs to validate the `==` and `!=` operators. Disable warning explicitly.

tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableArrayTests.cs
meziantou: Accepted. I restored the explicit null operator assertions in this test and kept both `array != null` and `null != array` checks. I also added a narrow `#pragma warning disable/restore MFAS0007` around those lines so we preserve operator coverage without changing analyzer settings globally.

tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableDictionaryTests.cs
meziantou: Accepted. I restored the explicit null operator assertions in this dictionary test so it continues validating the overloaded `==` and `!=` behavior directly. I also added a local `#pragma warning disable/restore MFAS0007` for just those assertions.

tests/Meziantou.Framework.Tests/Collections/ImmutableEquatableSetTests.cs
meziantou: Accepted. I restored the null operator assertions here as well so the set test still verifies both forms of the overloaded `!=` operator checks. I also scoped an explicit `#pragma warning disable/restore MFAS0007` to those lines only.

PR #1717, opened 2026-06-29
Add assertion analyzer rules for type-pattern readability

meziantou:
## Why
Using `Assert.True/False` with `is` / `is not` type checks is less expressive than dedicated assertion helpers. This change adds analyzer guidance and fixes so tests read more clearly and intent is explicit.

## What changed
- Added two new analyzer diagnostics:
  - `MFAS0012`: suggest `Assert.IsAssignableTo<T>(value)`
  - `MFAS0013`: suggest `Assert.IsNotAssignableTo<T>(value)`
- Extended assertion method selection analysis to detect and map these patterns:
  - `Assert.True(value is Type)` -> `Assert.IsAssignableTo<Type>(value)`
  - `Assert.False(value is Type)` -> `Assert.IsNotAssignableTo<Type>(value)`
  - `Assert.True(value is not Type)` -> `Assert.IsNotAssignableTo<Type>(value)`
  - `Assert.False(value is not Type)` -> `Assert.IsAssignableTo<Type>(value)`
- Updated the code fix provider to apply the corresponding generic assertion rewrite.
- Added analyzer/code-fix tests for all four transformations and updated analyzer rule documentation.

## Validation
- `dotnet run ./eng/update-all.cs`
- `dotnet test tests/Meziantou.Framework.Assertions.Analyzer.Tests/Meziantou.Framework.Assertions.Analyzer.Tests.csproj /p:TreatsWarningsAsErrors=true --nologo`

### DIFF
--- a/src/Meziantou.Framework.Assertions.Analyzer/Rules/AssertionMethodSelectionAnalyzerCommon.cs
+++ b/src/Meziantou.Framework.Assertions.Analyzer/Rules/AssertionMethodSelectionAnalyzerCommon.cs
@@ -81,22 +81,22 @@ internal static class AssertionMethodSelectionAnalyzerCommon
         }
 
         if (conditionOperation is IIsTypeOperation isTypeOperation &&
-            TryGetIsTypeOperationTypeSyntax(isTypeOperation, out var isTypeOperationTypeSyntax))
+            TryGetIsTypeOperationType(isTypeOperation, out var isTypeOperationType))
         {
             var assertionMethodName = conditionExpectedToBeFalse ? IsNotAssignableToAssertionMethodName : IsAssignableToAssertionMethodName;
-            match = new AssignableTypeCheckMatch(AssertionsAnalyzerHelpers.UnwrapImplicitConversion(isTypeOperation.ValueOperand), isTypeOperationTypeSyntax, assertionMethodName);
+            match = new AssignableTypeCheckMatch(AssertionsAnalyzerHelpers.UnwrapImplicitConversion(isTypeOperation.ValueOperand), isTypeOperationType, assertionMethodName);
             return true;
         }
 
         if (conditionOperation is not IIsPatternOperation isPatternOperation ||
-            !TryGetTypePattern(isPatternOperation.Pattern, out var typeSyntax, out var isNegated))
+            !TryGetTypePattern(isPatternOperation.Pattern, out var type, out var isNegated))
         {
             match = default;
             return false;
         }
 
         var patternAssertionMethodName = conditionExpectedToBeFalse == isNegated ? IsAssignableToAssertionMethodName : IsNotAssignableToAssertionMethodName;
-        match = new AssignableTypeCheckMatch(AssertionsAnalyzerHelpers.UnwrapImplicitConversion(isPatternOperation.Value), typeSyntax, patternAssertionMethodName);
+        match = new AssignableTypeCheckMatch(AssertionsAnalyzerHelpers.UnwrapImplicitConversion(isPatternOperation.Value), type, patternAssertionMethodName);
         return true;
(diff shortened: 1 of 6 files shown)


src/Meziantou.Framework.Assertions.Analyzer/Rules/AssertionMethodSelectionAnalyzerCommon.cs
```diff
@@ -193,9 +251,54 @@ private static bool IsNotNullPattern(IPatternOperation pattern)
...
+            })
+        {
+            typeSyntax = negatedTypeSyntax;
+            isNegated = true;
+            return true;
+        }
+
+        typeSyntax = null!;
+        isNegated = false;
+        return false;
```
meziantou: I'm sure you can use IOperation and not the SyntaxNode.

src/Meziantou.Framework.Assertions.Analyzer/Rules/AssertionMethodSelectionAnalyzerCommon.cs
```diff
@@ -193,9 +251,54 @@ private static bool IsNotNullPattern(IPatternOperation pattern)
...
+        }
+
+        if (isTypeOperation.TypeOperand is not null)
+        {
+            typeSyntax = SyntaxFactory.ParseTypeName(isTypeOperation.TypeOperand.ToDisplayString());
+            return true;
+        }
+
+        typeSyntax = null!;
+        return false;
```
meziantou, line 288: I'm sure you can use IOperation and not the SyntaxNode.

src/Meziantou.Framework.Assertions.Analyzer/Rules/AssertionMethodSelectionAnalyzerCommon.cs
meziantou: Good point, I updated this to use operation nodes directly. Type-pattern extraction now uses ITypePatternOperation/INegatedPatternOperation instead of syntax nodes.

src/Meziantou.Framework.Assertions.Analyzer/Rules/AssertionMethodSelectionAnalyzerCommon.cs
meziantou: Agreed, I switched this path to the operation model as well. The is-type branch now uses IIsTypeOperation.TypeOperand directly, without syntax parsing.

src/Meziantou.Framework.Assertions.Analyzer/Rules/AssertionMethodSelectionAnalyzerCommon.cs
```diff
@@ -193,9 +251,47 @@ private static bool IsNotNullPattern(IPatternOperation pattern)
...
+            {
+                Pattern: ITypePatternOperation { MatchedType: { } negatedType },
+            })
+        {
+            type = negatedType;
+            isNegated = true;
+            return true;
+        }
+
+        type = null!;
```
meziantou: No need for `!`. Annotate the parameter with `[NotNullWhen(true)]`
```
out [NotNullWhen(true)] ITypeSymbol? type
```

And apply a similar change to TryGetIsTypeOperationType and TryGetAssertCondition

meziantou: Great point, thanks. I updated TryGetTypePattern, TryGetIsTypeOperationType, and TryGetAssertCondition to use [NotNullWhen(true)] and removed the related null-forgiving assignments. This keeps null-state flow explicit without changing behavior.

PR #1727, opened 2026-06-30
Optimize collection snapshot materialization in assertion helpers

meziantou:
## Summary
- Reduce overhead in collection snapshot handling for synchronous and asynchronous assertions.
- Update assertion helpers to use the optimized snapshot flow consistently.
- Add coverage for the new snapshot behavior and formatting paths.

## Testing
- Not run (not requested)


src/Meziantou.Framework.Assertions/Utils/CollectionSnapshot.cs
```diff
@@ -3,74 +3,256 @@
...
+            if (_isComplete)
+            {
+                item = default!;
+                return false;
+            }
+
+            _enumerator ??= _source.GetEnumerator();
+
+            Debug.Assert(_enumerator is not null);
+            if (_enumerator.MoveNext())
```
meziantou: It only fetch one more item, whatever the `index` parameter. I think it should iterate to find the expected index.

PR #1747, opened 2026-07-05
Handle C# union types in the null-value analyzer

meziantou:
## Summary
- Treat C# union types as non-matching for the null-with-value-type analyzer helper
- Add a regression test to verify `Assert.NotNull` is not flagged for a union type

## Testing
- Added analyzer unit coverage for the union-type case

### DIFF
--- a/src/Meziantou.Framework.Assertions.Analyzer/AssertionsAnalyzerHelpers.cs
+++ b/src/Meziantou.Framework.Assertions.Analyzer/AssertionsAnalyzerHelpers.cs
@@ -6,6 +6,8 @@ namespace Meziantou.Framework.Analyzers.Assertions;
 internal static class AssertionsAnalyzerHelpers
 {
     public const string AssertMetadataName = "Meziantou.Framework.Assertions.Assert";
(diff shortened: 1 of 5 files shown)


src/Meziantou.Framework.Assertions.Analyzer/AssertionsAnalyzerHelpers.cs
```diff
@@ -37,11 +37,23 @@ public static bool IsValueType(ITypeSymbol? type)
...
     private static bool IsNullableValueType(ITypeSymbol type)
     {
         return type.OriginalDefinition.SpecialType == SpecialType.System_Nullable_T;
     }
+
+    private static bool IsCSharpUnionType(ITypeSymbol type)
+    {
+        foreach (var attribute in type.GetAttributes())
+        {
+            if (attribute.AttributeClass?.ToDisplayString() == "System.Runtime.CompilerServices.UnionAttribute")
```
meziantou: Compare types using ISymbol, not full name. Also, I think the type must also implement `IUnion`

PR #1843, opened 2026-08-16
Improve reliability of the TemporaryContainers tests

meziantou:
The `Meziantou.Framework.TemporaryContainers.Tests` tests fail intermittently on CI. Two recent examples, from the same run:

- [windows-2025, Release, InvariantGlobalization, shard 2](https://github.com/meziantou/Meziantou.Framework/actions/runs/31961857629/job/95200905860): all 8 `DockerContainerTests` tests failed at `docker build` / `docker create` with `ProcessExecutionException : Process exited with code 1. Command: C:\Windows\system32\docker.EXE`.
- [ubuntu-26.04, Debug, shard 2](https://github.com/meziantou/Meziantou.Framework/actions/runs/31961857629/job/95200905864): `PodmanContainerTests.Lifecycle_RestartStopAndDelete` failed at `RestartAsync` with podman exit code 125.

In both cases the report contains no usable information: the arguments are hidden by the default log verbosity, and the runtime standard error is dropped.

## Changes

**Report the container runtime output.** `ContainerCli` used `ProcessValidationMode.FailIfNonZeroExitCode`, which produces `Process exited with code 1. Command: docker.EXE` and discards everything the runtime wrote. It now validates the exit code itself and includes the sub-command, the exit code, the standard error and the standard output in the exception. `RunToStreamAsync` captures the standard error for the same reason. Only the sub-command is reported, not the whole argument list, because the create arguments contain environment variable values such as `MSSQL_SA_PASSWORD`. The exception type is unchanged, so the `OpenReadAsync` fallback chain still works.

The difference on a real failure:

```
The command '/usr/local/bin/container create' exited with code 1.
Standard error:
Error: interrupted: "XPC connection error: Connection invalid"
Ensure container system service has been started with `container system start`.
```

**Run one container test at a time per class.** The SDK generates `[assembly: Parallelization(Mode = All)]`, so the 8 tests of a runtime class each started their own container simultaneously, and both target frameworks did it concurrently. That is the mechanism behind image pull races and published port collisions. `XunitParallelizationMode` is now `Collections`: test classes still run in parallel, tests inside a class do not.

**Fix the start retry helper.** It was duplicated in five test classes, and `catch when (i < MaxRetries)` meant the last attempt never disposed its container, leaking one container per failing test. The shared `ContainerTestHelper` disposes every attempt, does not swallow the test cancellation, backs off progressively, and reports every attempt instead of the last one only. The two `MongoDbContainerTests` that called `StartAsync` without any retry now use it as well.

**Other hardening.** The HTTP polling is bounded by a deadline with a per-request timeout and retries on `IOException` / `SocketException` too; the reuse test cleanup is best-effort so it cannot mask the failure under test; the log streaming test reports the logs it received instead of a bare `Assert.True()` failure.

## Verification

Built with `/p:TreatWarningsAsErrors=true /p:RunAnalyzers=true` (0 warnings) and ran `dotnet run ./eng/update-all.cs` (no generated file changed). Full suite on net10.0 and net11.0: 67 passed, 17 skipped, 3 failed.

The 3 failures are pre-existing environment failures on the machine used (arm64 macOS, Apple `container` runtime), confirmed by re-running the same tests with the changes stashed: `SqlServerContainerTests.StartAsync_ConnectionStringWorks` (no arm64 mssql image) and the `MongoDbContainerTests` container tests (the mongo readiness log line never appears under that runtime). All Docker and Apple container integration tests pass.

## Note for the reviewer

Neither CI failure could be reproduced locally, so these changes remove the known amplifiers rather than a confirmed root cause. If the podman exit code 125 comes back, the failure report will now contain podman's actual error message.

### DIFF
--- a/src/Meziantou.Framework.TemporaryContainers/Internals/ContainerCli.cs
+++ b/src/Meziantou.Framework.TemporaryContainers/Internals/ContainerCli.cs
@@ -2,8 +2,6 @@ namespace Meziantou.Framework.TemporaryContainers.Internals;
 
 internal sealed class ContainerCli
 {
-    private const int MaxReportedOutputLength = 4000;
-
     private readonly string _executable;
 
     public ContainerCli(ContainerRuntime runtime, string executable)
@@ -18,7 +16,7 @@ internal sealed class ContainerCli
     {
         var wrapper = ProcessWrapper.Create(_executable)
             .WithArguments(args)
-            .WithValidation(ProcessValidationMode.None);
+            .WithValidation(allowNonZero ? ProcessValidationMode.None : ProcessValidationMode.FailIfNonZeroExitCode);
 
         if (input is not null)
             wrapper = wrapper.WithInputStream(input);
@@ -27,32 +25,17 @@ internal sealed class ContainerCli
 
         var standardOutput = string.Join('\n', result.Output.StandardOutput.Select(o => o.Text));
         var standardError = string.Join('\n', result.Output.StandardError.Select(o => o.Text));
-
-        if (!allowNonZero && !result.ExitCode.IsSuccess)
-            throw CreateExecutionException(args, result.ExitCode, standardOutput, standardError);
-
         return new CliResult(result.ExitCode.Value, standardOutput, standardError);
     }
 
     public async Task RunToStreamAsync(IReadOnlyList<string> args, Stream standardOutput, CancellationToken cancellationToken)
     {
-        var standardError = new StringBuilder();
-        var result = await ProcessWrapper.Create(_executable)
+        await ProcessWrapper.Create(_executable)
             .WithArguments(args)
-            .WithValidation(ProcessValidationMode.None)
+            .WithValidation(ProcessValidationMode.FailIfNonZeroExitCode)
             .WithOutputStream(OutputTarget.ToStream(standardOutput))
-            .WithErrorStream(OutputTarget.ToTextDelegate(line =>
-            {
-                lock (standardError)
-                {
-                    standardError.AppendLine(line);
-                }
-            }))
             .ExecuteAsync(cancellationToken)
             .ConfigureAwait(false);
-
-        if (!result.ExitCode.IsSuccess)
-            throw CreateExecutionException(args, result.ExitCode, standardOutput: "", standardError.ToString());
     }
 
     public ProcessInstance ExecuteStreaming(IReadOnlyList<string> args, Action<string> onStandardOutput, Action<string> onStandardError, CancellationToken cancellationToken)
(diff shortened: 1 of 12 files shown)


src/Meziantou.Framework.TemporaryContainers/Internals/ContainerCli.cs
```diff
@@ -2,6 +2,8 @@ namespace Meziantou.Framework.TemporaryContainers.Internals;
 
 internal sealed class ContainerCli
 {
+    private const int MaxReportedOutputLength = 4000;
```
meziantou: Revert all changes in this class

tests/Meziantou.Framework.TemporaryContainers.Tests/Meziantou.Framework.TemporaryContainers.Tests.csproj
```diff
@@ -2,6 +2,9 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(LatestTargetFrameworks)</TargetFrameworks>
+    <!-- Each test of a runtime test class starts its own container. Running them all at once saturates the CI agents and makes the
+         container runtimes fail transiently (image pull races, port collisions), so only one test per class runs at a time. -->
+    <XunitParallelizationMode>Collections</XunitParallelizationMode>
```
meziantou: Disable on test class directly, not globally

meziantou:
The first CI run on this branch reproduced the podman flake, and the new diagnostics identified it:

```
The command '/usr/bin/podman restart' exited with code 125.
Standard error:
time="..." level=warning msg="StopSignal SIGTERM failed to stop container wonderful_pascal in 10 seconds, resorting to SIGKILL"
Error: pasta failed with exit code 1:
netlink: Unexpected sequence number (0 != 42)
```

Restarting a container makes rootless podman tear down and rebuild the port forwarding, and `pasta` loses a netlink race while doing so. This is a podman/passt issue, not a defect of the library, and no test-side sequencing prevents it.

`Lifecycle_RestartStopAndDelete` is now retried from scratch, but only when the container runtime command line itself failed (`ProcessExecutionException`). An assertion failure is never retried, so a real regression of the restart, stop or delete behavior still fails on the first attempt.

src/Meziantou.Framework.TemporaryContainers/Internals/ContainerCli.cs
meziantou: Reverted in 5be1170ce — the class is back to its original content.

Heads-up on the consequence: the exception no longer carries the runtime standard error, so a failing command is described only by its exit code again. Concretely, the podman flake this PR is about reported `Process exited with code 1. Command: /usr/bin/podman` before, and `pasta failed with exit code 1: netlink: Unexpected sequence number` after. The retry added in `Lifecycle_RestartStopAndDelete` is unaffected — it keys off `ProcessExecutionException`, which `ProcessValidationMode.FailIfNonZeroExitCode` still throws.

Happy to propose the diagnostics separately (here or in `ProcessWrapper`) if you want it, in whatever shape suits you.

tests/Meziantou.Framework.TemporaryContainers.Tests/Meziantou.Framework.TemporaryContainers.Tests.csproj
meziantou: Done in 5be1170ce. The `XunitParallelizationMode` property is removed and each test class that starts a container now carries:

```csharp
[TestClass(DisableParallelization = true)]
```

on `DockerContainerTests`, `PodmanContainerTests`, `AppleContainerContainerTests`, `WslcContainerTests`, `RedisContainerTests`, `MongoDbContainerTests`, `PostgreSqlContainerTests` and `SqlServerContainerTests`. The assembly keeps the generated `ParallelMode.All`, which is the mode this attribute applies to, and the classes that do not start a container still run in parallel.

Verified rather than assumed: with the attribute, the 9 `DockerContainerTests` start/end times in the trx report have zero overlapping pairs, where they all overlapped before.

PR #1844, opened 2026-08-16
Add 24 assertion analyzer rules (MFAS0028-MFAS0051)

meziantou:
Compares the shipped `MFAS0001`–`MFAS0027` rules against the full `Assert` API surface in `ref/Meziantou.Framework.Assertions.cs` and closes three gaps: no correctness rules for async assertions, rule families that only handled `Assert.True`, and assertion methods that no rule pointed at (`InRange`, `IsNotType`, `ProperSubset`, `Fail`, …).

## Correctness rules (Error)

| Id | Detects |
| -- | -- |
| `MFAS0048` | The `Task` returned by an assertion is never awaited |
| `MFAS0049` | An un-awaited task compared against a plain value |
| `MFAS0050` | Assertion whose result is known in advance (`Assert.True(true)`, `Assert.Equal(x, x)`) |
| `MFAS0051` | `Assert.True(false)` → `Assert.Fail` (Warning) |

`MFAS0048` is the highest-value rule here. `CS4014` only fires inside an `async` method, so this got **no diagnostic at all** in a sync test method while the assertion silently never ran:

```csharp
[Fact]
public void Test()
{
    Assert.ThrowsAsync<InvalidOperationException>(() => FooAsync()); // never checked anything
}
```

`MFAS0049` covers `Assert.Equal(42, GetValueAsync())`, which compiles cleanly through the `Equal<TExpected, TActual>` overload (`OverloadResolutionPriority(-1)`) and can never succeed.

## `Assert.True`/`Assert.False` rewrites (`MFAS0031`–`MFAS0047`)

`==`, `Equals`/`object.Equals`/`string.Equals`, `SequenceEqual`, `ReferenceEquals`, inclusive range checks, `ISet<T>` proper subset/superset, negated conditions, and runtime type comparisons — each turning *"Assert.True() failed"* into a real value diff.

## Emptiness rules

`MFAS0028`/`MFAS0029` for `Any()` without a predicate (`MFAS0024`/`MFAS0025` deliberately match only the 2-parameter overload), and `MFAS0030` for count assertions against zero.

## Existing rules extended, without new ids

- `MFAS0006`/`MFAS0007` now handle `Assert.False` and `Nullable<T>.HasValue`; `TryGetNullCheckMatch` previously hardcoded `"True"`.
- The count analyzer now recognises `string.Length`, `ImmutableArray<T>.Length` and `Span<T>.Length`, handles `Assert.NotEqual`, and maps comparisons equivalent to an emptiness check onto `Empty`/`NotEmpty`.
- `readme.md` lists `DoesNotAll`, which exists in the API and is `MFAS0027`'s target.

## Reviewer notes

Three judgement calls worth a look:

1. **`MFAS0031` excludes floating point.** `double.NaN == double.NaN` is false but `Assert.Equal` uses `Equals`, where it is true, so rewriting would change behaviour. The rule is also limited to primitives, `string`, enums and `Nullable<T>` with no user-defined `operator ==`, so reference equality is never silently swapped for value equality. It also defers to `MFAS0004`/`MFAS0005` on count comparisons and to `MFAS0006`/`MFAS0007` on null comparisons.

2. **`MFAS0004` now covers `NotEmpty`**, so its title changed to "Use Assert.Empty or Assert.NotEmpty for zero count checks". Two existing tests shifted as a result: `Assert.False(x.Length == 0)` now fixes to `Assert.NotEmpty(x)` rather than `Assert.DoesNotHaveCount(0, x)`. I updated those tests rather than preserving the old output.

3. **`tests/Meziantou.Framework.Assertions.Tests/.editorconfig` disables `MFAS0050`/`MFAS0051`.** Those rules produced 44 errors in the library's own test suite — all genuine hits on code like `AssertionsAssert.True(false)` and `Assert.Equal(float.NaN, float.NaN, 0f)` that those tests must write in order to verify the assertions themselves. This follows the existing `tests/.editorconfig` precedent, but it is the file to revisit if you would rather narrow the rules instead.

One bug the tests caught: Roslyn includes omitted optional arguments in `IInvocationOperation.Arguments`, and their `Syntax` is the *whole invocation*, so the `Assert.Fail` fix initially produced `Assert.Fail(Assert.Fail(Assert.True(false)))`. It now filters on `ArgumentKind.Explicit`.

Every new rule ships a code fix, leaving `MFAS0003` as the only rule without one. The `MFAS0048`/`MFAS0049` fix converts a sync method to `async Task` when needed.

## Verification

- Analyzer tests: **168 passed**, 0 failed (up from 78 — 90 new, each rule with positive, negative and message-preservation cases)
- Assertions library tests: **776 passed**, 0 failed
- All 5 projects build clean with `/p:TreatWarningsAsErrors=true`
- `dotnet run ./eng/update-all.cs` run to regenerate the rule table; a second run is a no-op and `ref/` is unchanged (no public API change)


meziantou:
Split into a five-PR stack so each part can be reviewed on its own; superseded by #1845 → #1846 → #1847 → #1848 → #1849. The stack is also rebased onto current main, and includes four bug fixes found while hardening the rules afterwards.

PR #1849, opened 2026-08-16
Add assertion correctness rules (MFAS0048-MFAS0051)

meziantou:
**5 of 5**. Based on #1848. These are the highest-value rules in the stack: each catches a test that passes while asserting nothing.

| Id | Severity | Detects |
| -- | -- | -- |
| `MFAS0048` | Error | The `Task` returned by an assertion is discarded |
| `MFAS0049` | Error | A comparison assertion where exactly one side is an un-awaited task |
| `MFAS0050` | Error | An assertion whose result is known in advance |
| `MFAS0051` | Warning | `Assert.True(false)` → `Assert.Fail` |

`MFAS0048` is the one I would prioritise. `CS4014` only fires inside an `async` method, so this got **no diagnostic at all** while silently never running:

```csharp
[Fact]
public void Test()
{
    Assert.ThrowsAsync<InvalidOperationException>(() => FooAsync()); // never checked anything
}
```

It covers `ThrowsAsync`, `ThrowsAnyAsync`, `Throws(Func<Task>)`, `DoesNotThrow(Func<Task>)` and every `IAsyncEnumerable<T>` overload of `Equal`, `All`, `Empty`, `Single`, `Contains`, `HasCount*` and friends.

`MFAS0049` covers `Assert.Equal(42, GetValueAsync())`, which compiles through the `Equal<TExpected, TActual>` overload (`OverloadResolutionPriority(-1)`) and can never succeed. Comparing two tasks is left alone, since that may be intentional.

### Notes

- **The await fix targets the closest enclosing function**, not the method. It turns a `void` method into `async Task`, but offers **no fix** inside a non-async lambda or local function, where adding `async` would change the delegate type or a signature the caller relies on. Without this it produced `Action a = () => await ...`, which does not compile.
- **`MFAS0050` only considers simple references**, so `Assert.Equal(Next().Value, Next().Value)` is not reported — the instance a field is read from must itself be side-effect free.
- **`tests/Meziantou.Framework.Assertions.Tests/.editorconfig` disables `MFAS0050`/`MFAS0051`.** They produced 44 errors in the library's own test suite, all genuine hits on code like `AssertionsAssert.True(false)` that those tests must write in order to verify the assertions. This follows the existing `tests/.editorconfig` precedent, but it is the file to revisit if you would rather narrow the rules.

One bug the tests caught: Roslyn includes omitted optional arguments in `IInvocationOperation.Arguments`, and their `Syntax` is the *whole invocation*, so the `Assert.Fail` fix initially produced `Assert.Fail(Assert.Fail(Assert.True(false)))`. It now filters on `ArgumentKind.Explicit`.

### Verification

Analyzer tests 187 passed (from 78 at the start of the stack), library tests 786 passed, `/p:TreatWarningsAsErrors=true` clean, `eng/update-all.cs` run and idempotent.

### DIFF
--- a/tests/Meziantou.Framework.Assertions.Analyzer.Tests/AwaitAssertionRuleTests.cs
+++ b/tests/Meziantou.Framework.Assertions.Analyzer.Tests/AwaitAssertionRuleTests.cs
@@ -231,6 +231,34 @@ public sealed class AwaitAssertionRuleTests : AssertionsAnalyzerTestBase
     }
 
     [Fact]
+    public async Task Analyzer_ReportsDiagnostic_OnlyForAnExpressionStatement()
+    {
+        // Discarding the task or assigning it observes the result, so the assertion is not lost
+        var source = """
+            using System.Collections.Generic;
+            using System.Threading.Tasks;
+            using Meziantou.Framework.Assertions;
+
+            namespace Sample;
+
+            public static class TestClass
+            {
+                public static Task M(IAsyncEnumerable<int> actual)
+                {
+                    _ = Assert.Empty(actual);
+                    var task = Assert.Empty(actual);
+
+                    {|MFAS0048:Assert.Empty(actual)|};
+
+                    return task;
+                }
+            }
+            """;
+
+        await CreateAnalyzerTest<AwaitAssertionAnalyzerType>(source).RunAsync(XunitCancellationToken);
+    }
+
+    [Fact]
     public async Task Analyzer_DoesNotReportDiagnostic_WhenTaskIsConsumed()
     {
         var source = """
--- a/tests/Meziantou.Framework.Assertions.Analyzer.Tests/AwaitAssertionRuleTests.cs
+++ b/tests/Meziantou.Framework.Assertions.Analyzer.Tests/AwaitAssertionRuleTests.cs
@@ -245,6 +245,32 @@ public sealed class AwaitAssertionRuleTests : AssertionsAnalyzerTestBase
     }
 
     [Fact]
+    public async Task Analyzer_DoesNotReportDiagnostic_WhenTheAssertionReturnsAValueThatIsATask()
+    {
+        // Assert.Single<T> returns T, so these return a Task without the assertion being asynchronous
+        var source = """
+            using System.Collections.Generic;
+            using System.Threading.Tasks;
+            using Meziantou.Framework.Assertions;
+
+            namespace Sample;
+
+            public static class TestClass
+            {
+                public static void M(Dictionary<string, Task> tasksByName)
+                {
+                    Task[] items = [];
+                    Assert.Single(items);
+                    Assert.Single(items, item => item is not null);
+                    Assert.Contains("a", tasksByName);
+                }
+            }
+            """;
+
+        await CreateAnalyzerTest<AwaitAssertionAnalyzerType>(source).RunAsync(XunitCancellationToken);
+    }
+
+    [Fact]
     public async Task Analyzer_DoesNotReportDiagnostic_WhenTaskIsConsumed()
     {
         var source = """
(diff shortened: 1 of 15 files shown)


tests/Meziantou.Framework.Assertions.Analyzer.Tests/AwaitAssertionRuleTests.cs
```diff
@@ -0,0 +1,286 @@
+using AwaitAssertionAnalyzerType = Meziantou.Framework.Analyzers.Assertions.AwaitAssertionAnalyzer;
```
meziantou, line 1: Add a test for 

```
Task[] items = [];
Assert.Single(items); // Should not report event if Single return a Task
```

tests/Meziantou.Framework.Assertions.Analyzer.Tests/ConstantAssertionRuleTests.cs
```diff
@@ -0,0 +1,154 @@
...
+
+namespace Meziantou.Framework.Tests;
+
+public sealed class ConstantAssertionRuleTests : AssertionsAnalyzerTestBase
+{
+    [Theory]
+    [InlineData("Assert.True(true)")]
+    [InlineData("Assert.False(false)")]
+    [InlineData("Assert.Equal(value, value)")]
+    [InlineData("Assert.NotEqual(value, value)")]
```
meziantou: I don't think the rule should check Equal / NotEqual. Can you update the rule and the tests

tests/Meziantou.Framework.Assertions.Analyzer.Tests/AwaitAssertionRuleTests.cs
meziantou: Good catch, and it was broader than the one case. The rule tested the substituted return type, so any assertion whose generic return type happened to bind to `Task` was reported. Your example plus two more shapes were all false positives:

```csharp
Task[] items = [];
Assert.Single(items);                       // Single<T> returns T
Assert.Single(items, i => i is not null);   // same, predicate overload
Assert.Contains("a", tasksByName);          // Contains<TKey, TValue> returns TValue
```

It now tests `targetMethod.OriginalDefinition.ReturnType`, which is the type parameter for these and `Task`/`Task<T>` for the overloads that really do complete asynchronously. Added the test you asked for, covering all three shapes; it reports 3 diagnostics without the fix and none with it.

tests/Meziantou.Framework.Assertions.Analyzer.Tests/ConstantAssertionRuleTests.cs
meziantou: Done — the self-comparison check now covers only `Same`, `NotSame`, `Equivalent` and `NotEquivalent`.

Agreed on the reasoning too: `Equal`/`NotEqual` compare through `EqualityComparer<T>.Default` and therefore run the type's own `Equals`, so `Assert.Equal(x, x)` is a reflexivity assertion rather than a constant one — which is exactly what `SemanticVersionTests.Operator_DifferentValues` uses it for. `Same`/`NotSame` compare references and `Equivalent` compares structurally, so neither runs user code and both still report for any type.

This also let me delete the `HasBuiltInEquality` type restriction from the previous commit, which only existed to work around the `Equal` case, so the rule is simpler than before. I dropped the unordered variants along with `Equal`/`NotEqual` since they compare the same way — say the word if you would rather keep those.

tests/Meziantou.Framework.Assertions.Analyzer.Tests/AwaitAssertionRuleTests.cs
```diff
@@ -0,0 +1,312 @@
...
+            using System.Threading.Tasks;
+            using Meziantou.Framework.Assertions;
+
+            namespace Sample;
+
+            public static class TestClass
+            {
+                public static void M()
+                {
+                    Assert.Equal(42, {|MFAS0049:GetValueAsync()|});
```
meziantou: Only assertion return type should be evaluated by the rule. Assert code should not.

meziantou: Removed the rule entirely — the await rule now looks only at what the assertion returns, never at the code passed to it. That took out the descriptor, the argument matcher, its code-fix registration and its tests.

Kept a negative test so the boundary stays pinned:

```csharp
public static void TaskPassedAsAnArgument(Task<int> a, Task<int> b)
{
    // The rule looks at what the assertion returns, not at the code passed to it
    Assert.Equal(a, b);
    Assert.Equal(42, a);
}
```

One thing to confirm: since `MFAS0049` was never released, I closed the gap rather than leaving a hole in a new contiguous block, so the constant assertion rule is now `MFAS0049` and the `Assert.Fail` rule is `MFAS0050`. Happy to leave the number retired instead if you would rather never reuse an id.

PR #1850, opened 2026-08-16
Close coverage gaps in the FullPath analyzers

meziantou:
First of three PRs adding rules to `Meziantou.Framework.FullPath.Analyzer`. This one closes gaps in the rules that already exist; the next two add correctness rules and redundancy rules on top of it.

## The `.Value` blind spot

Every `Path.*` rule (MFFP0001, MFFP0003-MFFP0010) recognized a `FullPath` only when it reached the argument through the *implicit* conversion to `string`. These all went unreported:

```csharp
Path.GetExtension(fullPath.Value)
Path.GetExtension(fullPath.ToString())
Path.GetExtension((string)fullPath)
```

The unwrapping now lives in `FullPathAnalyzerCommon`, which is already shared with the code fix project, so the fixers keep pace with the analyzers instead of reporting a diagnostic no fix is offered for. Each fixer's private copy of `UnwrapImplicitConversion` is gone.

## `Path.Join`

MFFP0004 matched `Combine` only. `Path.Join(fullPath, "x")` slipped through and is arguably worse than `Combine`, since it does no rooting.

## Three declaration-site rules

MFFP0011 infers "this should be a `FullPath`" from return statements. The same inference applies at three more declaration sites:

| Id | Reports |
| -- | -- |
| `MFFP0012` | a get-only `string` property whose getter only returns `FullPath` values |
| `MFFP0013` | a `string` local that only ever holds `FullPath` values |
| `MFFP0014` | a `string` parameter whose call sites all pass a `FullPath` |

MFFP0014 only considers private methods and local functions, since every call site has to be visible in the compilation for the conclusion to hold. It also skips `params`, optional and by-ref parameters, and any method used as a delegate. It reports at compilation end, so it is a build-time diagnostic rather than a live one.

MFFP0013 leaves `var` declarations alone — there the developer asked for whatever the initializer produces — and bails out on any write that isn't a plain assignment.

## Fixes to MFFP0011

Two false positives, both surfaced while writing MFFP0012:

- it fired on property accessors, reporting `get_Foo` at the accessor. Accessors are now MFFP0012's job, reported at the property itself.
- it fired on members that override a base member or implement an interface, where the return type can't be changed.

## Notes

- The readme table is generated, but its regex stopped matching a descriptor at `isEnabledByDefault`, so MFFP0014 (which needs a `CompilationEnd` custom tag) was silently dropped from the table. `eng/update-all.cs` now allows trailing descriptor arguments.
- Analyzer tests: 27 -> 55, all passing. `dotnet run ./eng/update-all.cs` is clean.
- Dogfooding check: the repo builds every project that references `Meziantou.Framework.FullPath` with these analyzers. With the new rules temporarily raised to warnings, the repo produces zero hits — verified against a probe file to confirm the analyzers were actually running.

### DIFF
--- a/src/Meziantou.Framework.FullPath.Analyzer/ParameterShouldBeFullPathAnalyzer.cs
+++ b/src/Meziantou.Framework.FullPath.Analyzer/ParameterShouldBeFullPathAnalyzer.cs
@@ -1,0 +1,140 @@
+using System.Collections.Concurrent;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Meziantou.Framework.Analyzers.FullPath;
+
+/// <summary>
+/// Reports parameters that are declared as <see cref="string"/> but only ever receive <c>FullPath</c> arguments.
+/// </summary>
+/// <remarks>
+/// Only private methods and local functions are considered, because every call site must be visible in the
+/// compilation for the conclusion to hold.
+/// </remarks>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class ParameterShouldBeFullPathAnalyzer : DiagnosticAnalyzer
+{
+    public static readonly DiagnosticDescriptor Descriptor = new(
+        id: FullPathAnalyzerCommon.ParameterShouldBeFullPathDiagnosticId,
+        title: "Declare the parameter as FullPath instead of string",
+        messageFormat: "Parameter '{0}' only receives FullPath values and should be declared as FullPath instead of string",
+        category: "FullPath",
+        defaultSeverity: DiagnosticSeverity.Info,
+        isEnabledByDefault: true,
+        customTags: [WellKnownDiagnosticTags.CompilationEnd]);
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => [Descriptor];
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+        context.EnableConcurrentExecution();
+        context.RegisterCompilationStartAction(context =>
+        {
+            var analyzerContext = new FullPathContext(context.Compilation);
+            if (!analyzerContext.IsValid)
+                return;
+
+            var candidateCache = new ConcurrentDictionary<ISymbol, bool>(SymbolEqualityComparer.Default);
+            var arguments = new ConcurrentBag<ArgumentUsage>();
+            var excludedMethods = new ConcurrentBag<IMethodSymbol>();
+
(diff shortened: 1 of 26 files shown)


src/Meziantou.Framework.FullPath.Analyzer/ParameterShouldBeFullPathAnalyzer.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    private static bool IsCandidate(IMethodSymbol methodSymbol)
+    {
+        if (methodSymbol.MethodKind is not (MethodKind.Ordinary or MethodKind.LocalFunction))
+            return false;
+
+        // Every call site must be visible in the compilation
+        if (methodSymbol.MethodKind is MethodKind.Ordinary && methodSymbol.DeclaredAccessibility is not Accessibility.Private)
```
meziantou: I think you should introduce a new helper like https://github.com/meziantou/Meziantou.Analyzer/blob/a64ef7b041f7832b78fe502253277211791b67c4/src/Meziantou.Analyzer/Internals/SymbolExtensions.cs#L17-L31

Then call IsVisibleOutsideOfAssembly

meziantou: Done in 29545d548 — added `SymbolExtensions.IsVisibleOutsideOfAssembly` mirroring the Meziantou.Analyzer helper, and the candidate check is now just:

```csharp
// Every call site must be visible in the compilation
if (methodSymbol.IsVisibleOutsideOfAssembly())
    return false;
```

That also let the `MethodKind is MethodKind.Ordinary &&` guard go: a local function's `DeclaredAccessibility` isn't public/protected, so it falls out naturally.

Two things worth flagging from broadening this:

**It found a nullable parameter.** `GitTool.GetGitConfiguration(string? workingDirectory, string key)` started reporting. `FullPath` is a struct and can't represent null, so a parameter that explicitly admits null shouldn't be suggested for conversion — the rule now skips `NullableAnnotation.Annotated`. Good catch by your suggestion; that was a real hole.

**Two remaining hits I did not act on**, since they're your call rather than mechanical cleanup:

- `SnapshotUpdateStrategy.CanUpdateSnapshotInternal(SnapshotSettings, string path, ...)` (reported twice — the file is linked into both SnapshotTesting and InlineSnapshotTesting). Every call site passes a `FullPath`, and changing it does compile: the body forwards `path` to the `public abstract CanUpdateSnapshot(..., string path, ...)`, where the implicit conversion applies. But it changes the shape of an internal API next to a public abstract one, so I left it. Happy to change it if you want.

I did fix the third one, a private `CreateSymlink(string source, ...)` test helper in `FullPathTests` where all eight call sites pass a `FullPath`. Only `source` was reported — `target` receives string literals too — which is a nice precision check on the rule.

Analyzer tests 55 -> 59.

meziantou: Change all internal methods if the analyzer reports them. Do not update public methods.

PR #1851, opened 2026-08-16
Add FullPath correctness rules

meziantou:
Second of three PRs adding rules to `Meziantou.Framework.FullPath.Analyzer`.

> **Stacked on #1850** — the base branch is `claude/fullpath-analyzer-coverage-gaps`, so the diff stays scoped to this group. Retarget to `main` once #1850 merges.

The existing rules all report a missed simplification: the code is correct, it just isn't idiomatic. These two report a **wrong result**, so they are warnings rather than suggestions. Both exploit the implicit `FullPath` -> `string` conversion, and both compile without a squiggle today.

## MFFP0015 — comparing a `FullPath` to a `string`

```csharp
if (fullPath == "C:\\Temp\\file.txt")   // binds to string.operator ==
```

`operator ==(FullPath, FullPath)` isn't applicable, since there is no implicit conversion from `string` to `FullPath`. Overload resolution converts the left side and picks `string.operator ==`, which is **ordinal** — while `FullPathComparer.Default` is **case-insensitive on Windows**. Also covers `fullPath.Value == other.Value` and the `string.Equals` / `string.Compare` / `CompareTo` forms.

Passing an explicit `StringComparison` is not reported: that is the escape hatch for deliberately comparing as strings.

## MFFP0016 — `fullPath.Equals(someString)` is always `false`

A `string` argument makes the call bind to `FullPath.Equals(object)`, whose body is `obj is FullPath path && Equals(path)` ([FullPath.cs:174](https://github.com/meziantou/Meziantou.Framework/blob/main/src/Meziantou.Framework.FullPath/FullPath.cs#L174)). A `string` is never a `FullPath`, so the branch is dead.

An `object` argument is deliberately not reported — it may hold a boxed `FullPath` at run time.

## Notes

- MFFP0017 (`IsChildOf` instead of `StartsWith`) was dropped at review request. `MFFP0017` is currently unused; #1852 still starts at `MFFP0018`, leaving a gap. Say the word if you would rather I renumber #1852 down to `MFFP0017`-`MFFP0022`.
- No code fixes. For MFFP0015 the right fix depends on whether the other side should become a `FullPath` or the comparison should become explicitly ordinal, so it isn't something to apply unattended.
- Analyzer tests: 59 -> 73, all passing. `dotnet run ./eng/update-all.cs` is clean.
- Because these are warnings and the repo builds every `FullPath`-referencing project with these analyzers under `TreatsWarningsAsErrors`, a single real hit would break the build. The repo produces zero hits; verified against a probe file that both rules do fire in a dogfooding build.


tests/Meziantou.Framework.FullPath.Analyzer.Tests/CompareFullPathAsStringRuleTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                    public static bool M(FullPath fullPath, FullPath other, string text)
+                    {
+                        return {|MFFP0015:{{expression.TrimEnd()}}|};
+                    }
+                }
+            }
+            """;
+
+        await CreateAnalyzerTest<CompareFullPathAsStringAnalyzerType>(source).RunAsync(XunitCancellationToken);
+    }
```
meziantou, line 31:
```suggestion
    [Theory]
    [InlineData("fullPath == \"value\"")]
    [InlineData("fullPath != \"value\"")]
    [InlineData("fullPath.Value == other.Value")]
    [InlineData("fullPath.ToString() == text")]
    [InlineData("text == fullPath")]
    public async Task Analyzer_ReportDiagnostic_ForStringComparison(string expression)
    {
        var source = $$"""
            using Meziantou.Framework;

            namespace Sample
            {
                public static class TestClass
                {
                    public static bool M(FullPath fullPath, FullPath other, string text)
                    {
                        return {|MFFP0015:{{expression.TrimEnd()}}|};
                    }
                }
            }
            """;

        await CreateAnalyzerTest<CompareFullPathAsStringAnalyzerType>(source).RunAsync(XunitCancellationToken);
    }
```

tests/Meziantou.Framework.FullPath.Analyzer.Tests/CompareFullPathAsStringRuleTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        var source = $$"""
+            using Meziantou.Framework;
+
+            namespace Sample
+            {
+                public static class TestClass
+                {
+                    public static bool M(FullPath fullPath, FullPath other, string text)
+                    {
+                        return {|MFFP0015:{{expression.TrimEnd()}}|};
```
meziantou:
```suggestion
                        return {|MFFP0015:{{expression}}|};
```

tests/Meziantou.Framework.FullPath.Analyzer.Tests/StartsWithInsteadOfIsChildOfRuleTests.cs
```diff
@@ -0,0 +1,91 @@
+using StartsWithInsteadOfIsChildOfAnalyzerType = Meziantou.Framework.Analyzers.FullPath.StartsWithInsteadOfIsChildOfAnalyzer;
+
+namespace Meziantou.Framework.Tests;
+
+public sealed class StartsWithInsteadOfIsChildOfRuleTests : FullPathAnalyzerTestBase
```
meziantou: I don't think this rule is useful. Can you remove the rule and its tests

meziantou: Removed in 080cc93bf — the analyzer, its tests, the diagnostic id constant, and the generated readme row are all gone. Analyzer tests 77 -> 73.

One knock-on: #1852 starts at `MFFP0018`, so `MFFP0017` is now an unused id. I left the gap rather than renumbering, since that would rewrite a PR you're already reviewing — but nothing is released yet, so renumbering #1852 down to `MFFP0017`-`MFFP0022` is a one-command change if you'd prefer contiguous ids. Your call.

PR #1857, opened 2026-08-17
Add SnapshotTesting.SkiaSharp package

meziantou:
The SkiaSharp counterpart of `Meziantou.Framework.SnapshotTesting.ImageSharp`. `AddSkiaSharp()` registers a serializer for `SKImage`, `SKBitmap`, `SKPixmap` and `SKSurface`, a comparer, and an `SKColor` converter for the human-readable serializer.

```csharp
SnapshotSettings.Default.AddSkiaSharp();

using var bitmap = SKBitmap.Decode("sample.png");
Snapshot.Validate(bitmap, SnapshotType.Png);
```

## What it does

The serializer encodes PNG, JPEG and WebP — the three formats Skia's encoder supports. Any other `SnapshotType` returns `false` so another serializer can handle it.

The comparer decodes both snapshots through `SKCodec` into a tightly packed `Rgba8888`/`Unpremul` bitmap, so two snapshots of the same image compare equal whatever format each was stored in. It then compares the pixels exactly, or computes the mean SSIM across R, G and B when `ImageComparisonSettings.SimilarityThreshold` is set.

Skia hands the pixels over as one contiguous buffer rather than per row, so the `Vector512`/`Vector256`/`Vector128` loops run over a single flat span instead of the per-row iteration the ImageSharp comparer uses.

## Deliberate differences from the ImageSharp package

Both follow what Skia actually supports:

- **No TIFF.** Skia neither encodes nor decodes it, so registering a comparer for it would only turn every TIFF comparison into a failure.
- **GIF and ICO get a comparer** even though the serializer cannot produce them, because Skia decodes both and a caller may still snapshot such bytes from elsewhere.

I verified this decode matrix against the image assets in `Meziantou.Framework.SnapshotTesting.Tests` rather than assuming it — BMP, GIF, ICO, PNG and JPEG all round-trip; every TIFF asset fails to decode.

## For reviewers

- SkiaSharp ships its native libraries for Windows and macOS in the main package; the readme points Linux consumers at `SkiaSharp.NativeAssets.Linux`.
- Like the ImageSharp package, this one has no test project. Verification was done out-of-tree:
  - All four input types serialize; unsupported types and values return `false`; exact comparison distinguishes identical, different, and differently-sized images; PNG vs WebP of the same bitmap compares equal.
  - SSIM parity against the ImageSharp implementation on identical PNG bytes, across sizes hitting every vector width and the scalar remainder: agreement to float precision (max delta 1.8e-7).
- `dotnet run ./eng/update-all.cs` was run to convergence — it added the README row, the solution entry and `slnx/`, and a second run reports no changes. `ref/` was generated by the build.
- `Meziantou.Framework.SnapshotTesting.Tests` passes 138/138 on net11.0.
- `eng/validate-nuget.cs` reports errors 112 (non-deterministic paths) and 119 (source not accessible) for the new package, but packing the existing ImageSharp package locally produces exactly the same two error classes — both come from a local, unpushed build rather than a CI build.


meziantou:
Followed up in b4514ad81: the three SSIM implementations are now one.

**Why it collapses so cleanly.** None of the three actually cares which channel is which — all read bytes 0, 1 and 2 of each packed pixel and ignore byte 3. The built-in comparer does that over ARGB (bytes 2, 1, 0 = R, G, B), the other two over RGBA (bytes 0, 1, 2 = R, G, B). Since the score is the *unweighted mean* of the three per-channel scores, the naming cancels out. Verified on images with deliberately lopsided channel distributions, so a mix-up would show.

**Shape.** `SsimAccumulator` lives in the core project and is linked into the two satellites as a `Compile` item — the pattern the core csproj already uses for `StringUtils.cs` and `ExecutableFinder.cs`. Deliberately *not* `InternalsVisibleTo`: the three packages version independently, so a cross-assembly call would bind to something outside the package contract. It is an accumulator rather than a function because ImageSharp does not guarantee an image sits in one buffer — `ProcessPixelRows` hands out a row at a time. The built-in and SkiaSharp comparers add a whole image in one call; ImageSharp adds a row at a time.

**One real bug found on the way.** The `float` lane accumulators only represent integers exactly up to 2^24, and the largest per-pixel term is a channel product (≤ 65025), so a long enough run silently loses the low bits. Accumulating a 256×256 image in one go drifted 2.8e-4 from the exact value; a 4K image would have lost the increments entirely. `Add` now folds into its `double` fields every 256 pixels, which keeps every value reached — in a lane, and in every partial sum of the horizontal reduction — under 2^24.

The three comparers now return bit-identical scores, all equal to a plain double-precision reference:

```
        before                                   after
 17x3   .607207686 .607207686 .607207686   →  all three identical
 64x48  .863324732 .863255709 .863324732   →  all three identical
129x7   .747727424 .747729212 .747727424   →  all three identical
256x256 .884874076 .885151178 .884874076   →  all three identical
        builtin    skia       imagesharp
```

Net −421/+356 lines. Two tests cover the two properties (split-independence, and agreement with a double-precision computation); both fail if the fold is removed. Suite is 144/144 on net11.0.

Unrelated, spotted while in there: `Meziantou.Framework.SnapshotTesting.ImageSharp` has a `System.Numerics.Tensors` PackageReference that nothing uses — it predates this PR, and dropping it changes the published dependency set, so I left it alone.

PR #1895, opened 2026-08-20
Add generated code detection helpers to Meziantou.Framework.Roslyn

meziantou:
## What

Adds `GeneratedCodeExtensions` to `Meziantou.Framework.Roslyn`, a port of Roslyn's internal [`Roslyn.Utilities.GeneratedCodeUtilities`](https://github.com/dotnet/roslyn/blob/2f1cb17a6c23f1700955937641c39d1785173d71/src/Compilers/Core/Portable/SourceGeneration/GeneratedCodeUtilities.cs) (permalink pinned in the file header).

| Member | Behavior |
|---|---|
| `SyntaxTree.IsGeneratedCode(CancellationToken)` | file-name heuristic + `<auto-generated>` leading comment |
| `SyntaxTree.IsGeneratedCode(AnalyzerOptions \| AnalyzerConfigOptionsProvider, CancellationToken)` | `generated_code = true\|false` from `.editorconfig` wins, falls back to the heuristics |
| `GeneratedCodeExtensions.IsGeneratedCodeFile(string?)` | `TemporaryGeneratedFile_*`, `*.designer.*`, `*.generated.*`, `*.g.*`, `*.g.i.*` |
| `SyntaxNode.HasAutoGeneratedComment()` | leading `//` / `/* */` trivia containing `<auto-generated` or `<autogenerated` |

## Why

Roslyn knows how to answer this question, but the logic is not public. Analyzers get it indirectly through `context.ConfigureGeneratedCodeAnalysis(...)` — source generators, diagnostic suppressors, and code fixes have no equivalent and must reimplement the heuristics, which is why analyzer packages such as `Microsoft.CodeAnalysis.BannedApiAnalyzers` vendor a copy of that Roslyn file.

## Notes for reviewers

- **The path parsing deliberately does not use `System.IO.Path`.** Roslyn goes through `FileNameUtilities`, which treats `\`, `/` and `:` as separators on *every* platform. `Path.GetFileName(@"Generator\Ns\Hint.g.cs")` returns the whole string on Linux/macOS, so a generator hint path would be misclassified there. Tests cover both separator styles.
- `Sample.g.cs` is generated, `Sample.g` is not — Roslyn only applies the suffix rules when the path has an extension.
- Documentation comments are not comments for this purpose, so `/// <auto-generated/>` does not trigger detection. This matches `CSharpCompilation.CreateAnalyzerDriver`, which only passes `SingleLineCommentTrivia` / `MultiLineCommentTrivia`.
- The `generated_code` option precedence (option first, heuristics as fallback) mirrors `AnalyzerDriver.IsGeneratedCode`.
- `HasAutoGeneratedComment` uses the ordinal `string.Contains(string)` with a scoped `CA1307`/`MA0074` pragma: the `StringComparison` overload does not exist on netstandard2.0, which is the package's real target. Follows the existing `CA1018` precedent in `EmbeddedAttribute.cs`.
- An `ISymbol.IsGeneratedCode` helper based on `[GeneratedCode]` was considered and dropped — its semantics are different enough from the syntax-tree checks to be confusing under the same name.
- `RoslynPackageTests.cs` hard-codes the package contents, so its `SourceFileNames` list and entry assertions are updated too.

## Verification

- Full test suite on Roslyn 4.8 / 5.0 / 5.3 / 5.6 × net10.0 + net11.0: **210 passed each, 0 failed** (was 134; the test projects compile the sources with `MEZIANTOU_FRAMEWORK_ROSLYN_ENABLE_WARNINGS` and `/p:TreatsWarningsAsErrors=true`).
- Package tests: **14 passed** — includes a real pack and a consumer build with warnings as errors.
- `dotnet run ./eng/update-all.cs`: no generated changes (this project sets `IncludeBuildOutput=false` and has no `ref/` baseline).
- Rebuilt the six in-repo projects that compile these sources directly (`FullPath.Analyzer`, `FullPath.CodeFix`, `FastEnumGenerator`, `Yaml.SourceGenerator`, `StronglyTypedId`, and the package project): all succeeded.


src/Meziantou.Framework.Roslyn/GeneratedCodeExtensions.cs
```diff
@@ -0,0 +1,142 @@
...
+                return true;
+#pragma warning restore MA0074
+#pragma warning restore CA1307
+        }
+
+        return false;
+    }
+
+    // Roslyn considers '\', '/' and ':' as separators on every platform, so the result doesn't depend on the current OS
+    private static string GetFileName(string path, bool includeExtension)
```
meziantou: Could it accept a ReadOnlySpan and return a ReadOnlySpan

meziantou: Done in ef1c6b6df. `GetFileName` now takes and returns a `ReadOnlySpan<char>`, so `IsGeneratedCodeFile` is allocation-free — the two `Substring` calls are gone.

I also dropped the `includeExtension` parameter: the caller already computes the extension index for the "does it have an extension" check, so it just slices with it instead of re-scanning.

`System.Memory` is a transitive dependency of `Microsoft.CodeAnalysis.Common`, so `ReadOnlySpan<char>`, `AsSpan()` and the `StringComparison` overloads of `StartsWith`/`EndsWith` are available to netstandard2.0 consumers. Verified by building the package project and all six in-repo consumers (the analyzers target netstandard2.0), plus the full test suite on Roslyn 4.8/5.0/5.3/5.6 and the package tests.

PR #1900, opened 2026-08-20
Move ConcurrentObservableCollection to Meziantou.Framework.Threading

meziantou:
Moves `ConcurrentObservableCollection<T>` out of `Meziantou.Framework.WPF` and into `Meziantou.Framework.Threading`, and bumps the WPF package to `4.0.0` for the breaking change.

## Why

Nothing about a thread-safe bindable collection is WPF specific — it lived in the WPF package only because it relied on `Dispatcher`. It now sits in `Meziantou.Framework.Collections.Concurrent`, alongside `ConcurrentHashSet<T>` and `SynchronizedList<T>`.

## What changed

`ConcurrentObservableCollection<T>`, `IReadOnlyObservableCollection<T>` and the six supporting internal types moved (via `git mv`, so history follows) and changed namespace from `Meziantou.Framework.WPF.Collections` to `Meziantou.Framework.Collections.Concurrent`.

Since the Threading package is cross-platform, `Dispatcher` is replaced by `SynchronizationContext`:

- `ConcurrentObservableCollection(Dispatcher)` → `ConcurrentObservableCollection(SynchronizationContext)`
- `dispatcher.BeginInvoke(…)` → `_synchronizationContext.Post(…)`

Creating the collection on the UI thread keeps the previous behavior, since the ambient context is then the dispatcher's.

## This is a hard break for Meziantou.Framework.WPF

The `Meziantou.Framework.WPF.Collections` namespace is **gone** — no obsolete shim, no type forward. The WPF package does not reference `Meziantou.Framework.Threading` either, so the type does not reach WPF consumers transitively. Migration is: add the `Meziantou.Framework.Threading` package, change the `using` to `Meziantou.Framework.Collections.Concurrent`, and construct the collection on the UI thread instead of passing a `Dispatcher`.

WPF goes to `4.0.0`; Threading stays at `3.0.0` since it only gains a type. The WPF package description drops its mention of collections, and the collection sample now references the Threading project directly since it no longer gets the type through the WPF project.

## Two decisions worth a look

**The base type is no longer sealed, and exposes `protected internal virtual bool IsOnSynchronizationContextThread()`.** The default implementation compares `SynchronizationContext.Current` against the instance the collection was built with. A derived type bound to a *thread* rather than to a context instance can override it — relevant because WPF's UI-thread `SynchronizationContext` is a different instance from `new DispatcherSynchronizationContext(dispatcher)`, so a plain reference check would fall back to posting and mutations made on the UI thread would stop being visible in `AsObservable` immediately. Nothing in the repo overrides it now that the shim is gone; happy to drop the hook if you would rather keep the type sealed.

**The parameterless constructor degrades instead of throwing** when the current thread has no context — it falls back to a default `SynchronizationContext`. Throwing would have broken `JsonSerializer.Deserialize<ConcurrentObservableCollection<T>>`, which needs that constructor. This isn't a regression: `Dispatcher.CurrentDispatcher` on a thread without a message pump queued the notifications forever. It's documented on the constructor.

## Tests

All 30 existing cases moved to `Meziantou.Framework.Threading.Tests`. The `MemberData` had to become lazy (a `CollectionKind` enum) because it constructed the collections on the data-enumeration thread, which no longer matches the thread running the test. Added two tests for the genuinely new behavior: cross-thread mutations queueing on the context, and the off-thread access guard.

## Verification

Rebased onto `main` (the conflict was in the generated `ref/Meziantou.Framework.Threading.cs`, where #1899's `SynchronizedList` change landed next to the inserted block — resolved by regenerating the file, so it carries both).

All 7 affected projects build with `-p:TreatWarningsAsErrors=true`, zero diagnostics. Threading tests: **262 passed, 0 failed, 0 skipped** (131 × net10.0/net11.0). `dotnet run ./eng/update-all.cs` runs to a clean exit; `ref/*.cs` is the regenerated output.

⚠️ **`Meziantou.Framework.WPF.Tests` was not run.** This was developed on macOS, where the `net*-windows` test host needs `Microsoft.WindowsDesktop.App`, which isn't available — CI already excludes that project on non-Windows via `eng/build.platform-exclusions.proj`. It compiles cleanly. With the shim gone the WPF project no longer contains any of this code, so the untested surface is now limited to the collection sample building against the moved type.


src/Meziantou.Framework.Threading/Meziantou.Framework.Threading.csproj
meziantou:
```suggestion
    <Version>3.1.0</Version>
```

PR #1997, opened 2026-08-26
Strip the ALL_HEADERS block from SQL batch command text

meziantou:
A TDS 7.2+ SQLBatch payload is `ALL_HEADERS` followed by the SQL text. The RPC path skipped the header block; the SQLBatch path decoded the *entire* payload — header included — as UTF-16 and exposed it as `TdsQueryContext.CommandText`.

Verified against a real `SqlConnection`: sending `SELECT 1` produced a **19-character** `CommandText` whose first 11 code points are `0016 0000 0012 0000 0002 0000 0000 0000 0000 0001 0000`, followed by `SELECT 1`. `CommandText.StartsWith("SELECT")` was `false`.

`CommandText` is a documented public API — the readme shows it as `"SELECT * FROM Users WHERE Id = @id"`. Any callback doing `StartsWith`, `Equals`, an anchored regex, or logging the statement got a wrong answer or corrupted output. The built-in query engine survived only because the SQL parser happens to tolerate leading control characters.

This extracts the header-skipping into `GetAllHeadersLength` and uses it on both paths. The shared helper also validates the block structure (each header's length must fit and the headers must add up exactly to the declared total) instead of trusting the first `DWORD`, so a TDS 7.1 client that sends no header block at all is still handled correctly. That is a small improvement to the RPC path too.

The existing `SqlClient_TextQuery_WithoutParameters_UsesSqlBatch` test asserted `Contains(marker)`, which is exactly why this went unnoticed — it is now an equality assertion, and a new test covers the round-trip explicitly.

Found by a review of `src/Meziantou.Framework.TdsServer`.

### DIFF
--- a/src/Meziantou.Framework.TdsServer/Protocol/TdsQueryRequestParser.cs
+++ b/src/Meziantou.Framework.TdsServer/Protocol/TdsQueryRequestParser.cs
@@ -18,7 +18,7 @@ internal static class TdsQueryRequestParser
             {
                 RemoteEndPoint = remoteEndPoint,
                 RequestType = TdsQueryRequestType.SqlBatch,
-                CommandText = DecodeUnicode(packet.Payload),
+                CommandText = DecodeUnicode(packet.Payload.AsSpan(GetAllHeadersLength(packet.Payload))),
                 UserContext = userContext,
             },
             TdsPacketType.Rpc => CreateRpcContext(packet.Payload, remoteEndPoint, userContext),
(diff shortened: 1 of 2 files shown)


meziantou:
Closing as a duplicate of #1991, which was opened first and covers the same finding.

The two converged on the same design independently — shared `GetAllHeadersLength`/`GetPayloadOffsetAfterAllHeaders` helper used by both the SQLBatch and RPC paths, with the header block structurally validated (each header's length must fit, and the headers must tile the block exactly) rather than trusting the leading DWORD, so TDS 7.1 clients that send no block still work.

#1991 is a strict superset on the test side: it tightens two loose `Contains` assertions to equality where this PR tightened one, and adds a no-control-characters assertion. Nothing here is worth porting over.

PR #2031, opened 2026-08-26
Return an ImmutableArray from GetAllInterfacesIncludingSelf

### LINKED ISSUE #2023: Meziantou.Framework.Roslyn: GetAllInterfacesIncludingSelf returns an IList<T> that throws on mutation for non-interfaces
[`TypeSymbolExtensions.GetAllInterfacesIncludingSelf`](https://github.com/meziantou/Meziantou.Framework/blob/abd00ecc7/src/Meziantou.Framework.Roslyn/TypeSymbolExtensions.cs#L18) returns `IList<INamedTypeSymbol>`, but the concrete type depends on the input:

- interface symbol -> a fresh `List<INamedTypeSymbol>` (mutable)
- anything else -> `type.AllInterfaces`, an `ImmutableArray<T>` boxed into `IList<T>`, whose `IList` implementation throws

Verified:

```
class     -> ImmutableArray`1 / Add throws NotSupportedException
interface -> List`1          / Add succeeds
```

An `IList<T>` return type advertises mutability. A consumer who adds to the result will pass tests written against an interface symbol and crash at runtime on a class symbol. The boxing also allocates on the hot path of a helper likely to be called per-symbol.

### Secondary: symbol comparison

The membership test on [line 21](https://github.com/meziantou/Meziantou.Framework/blob/abd00ecc7/src/Meziantou.Framework.Roslyn/TypeSymbolExtensions.cs#L21) uses `ImmutableArray.Contains`, which falls back to `EqualityComparer<INamedTypeSymbol>.Default` rather than the `SymbolEqualityComparer.Default` used by every other comparison in the project ([`TypeSymbolExtensions.cs#L294`](https://github.com/meziantou/Meziantou.Framework/blob/abd00ecc7/src/Meziantou.Framework.Roslyn/TypeSymbolExtensions.cs#L294), [`MethodSymbolExtensions.cs#L165`](https://github.com/meziantou/Meziantou.Framework/blob/abd00ecc7/src/Meziantou.Framework.Roslyn/MethodSymbolExtensions.cs#L165), [`LocalDataFlowAnalysis.cs#L380`](https://github.com/meziantou/Meziantou.Framework/blob/abd00ecc7/src/Meziantou.Framework.Roslyn/LocalDataFlowAnalysis.cs#L380)). This is what `RS1024` exists to flag; the file's blanket `#pragma warning disable` hides it.

### Suggested fix

Return `ImmutableArray<INamedTypeSymbol>` and build the interface case with `allInterfaces.Add(namedType)` (a cheap immutable append). That is honest about mutability, avoids the boxing, and keeps the common path allocation-free. Switch line 21 to an explicit `SymbolEqualityComparer.Default` loop, or `allInterfaces.Contains(namedType, SymbolEqualityComparer.Default)`.

This is a source-breaking change for existing callers, so it wants a version bump. `IReadOnlyList<INamedTypeSymbol>` is the lower-friction alternative if that matters.

<sub>Found during a review of `src/Meziantou.Framework.Roslyn` at `abd00ecc7`.</sub>

meziantou:
Fixes #2023

## What changed

`TypeSymbolExtensions.GetAllInterfacesIncludingSelf` now returns `ImmutableArray<INamedTypeSymbol>` instead of `IList<INamedTypeSymbol>`.

## Why

The old signature advertised mutability, but the concrete type depended on the input:

- interface symbol → a fresh, mutable `List<INamedTypeSymbol>`
- anything else → `type.AllInterfaces`, an `ImmutableArray<T>` boxed into `IList<T>`, whose mutating members throw `NotSupportedException`

A caller that added to the result would pass tests written against an interface symbol and crash at runtime on a class symbol. The boxing also allocated on the hot path of a helper likely to be called per-symbol.

Returning `ImmutableArray<INamedTypeSymbol>` is honest about mutability, removes the boxing, and lets the interface case use a single cheap `allInterfaces.Add(namedType)` append instead of allocating and copying into a `List<T>`.

## Symbol comparison

The membership test used `ImmutableArray.Contains`, which falls back to `EqualityComparer<INamedTypeSymbol>.Default` rather than the `SymbolEqualityComparer.Default` used by every other comparison in the project. It is now an explicit `SymbolEqualityComparer.Default` loop — allocation-free, and consistent with the rest of the codebase. This is what RS1024 exists to flag; the file's blanket `#pragma warning disable` was hiding it.

## Notes for reviewers

- **This is source-breaking** for consumers of the source package. `<Version>` is left at `1.0.4` because `eng/update-versions.cs` bumps versions from commit history in a separate PR (as in 7f1dae0d4). Worth a **minor** bump rather than a patch, which the script may not infer on its own.
- The only in-repo caller is the test suite, updated here.
- `IReadOnlyList<INamedTypeSymbol>` was the lower-friction alternative mentioned in the issue, but it keeps the boxing allocation, so this takes the `ImmutableArray` route the issue recommends.

## Tests

Updated the existing test to compare with `SymbolEqualityComparer.Default`, and added two cases: a class symbol returns its interfaces without including itself, and an interface with a base interface is not duplicated in the result.

## Verification

- Built `src/Meziantou.Framework.Roslyn` (net11.0 + netstandard2.0) and all four Roslyn test variants with `/p:TreatsWarningsAsErrors=true` — 0 warnings, 0 errors.
- Full test suite passes on Roslyn 4.8, 5.0, 5.3 and 5.6 — 214/214 each, across net10.0 and net11.0.
- `dotnet run ./eng/update-all.cs` ran clean and produced no additional file changes.

### DIFF
--- a/src/Meziantou.Framework.Roslyn/TypeSymbolExtensions.cs
+++ b/src/Meziantou.Framework.Roslyn/TypeSymbolExtensions.cs
@@ -4,6 +4,7 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
 using Microsoft.CodeAnalysis;
 
@@ -15,18 +16,19 @@ namespace Meziantou.Framework.Roslyn;
 #endif
 internal static partial class TypeSymbolExtensions
 {
-    public static IList<INamedTypeSymbol> GetAllInterfacesIncludingSelf(this ITypeSymbol type)
+    public static ImmutableArray<INamedTypeSymbol> GetAllInterfacesIncludingSelf(this ITypeSymbol type)
     {
         var allInterfaces = type.AllInterfaces;
-        if (type is INamedTypeSymbol namedType && namedType.TypeKind == TypeKind.Interface && !allInterfaces.Contains(namedType))
+        if (type is not INamedTypeSymbol { TypeKind: TypeKind.Interface } namedType)
+            return allInterfaces;
+
+        foreach (var @interface in allInterfaces)
         {
-            var result = new List<INamedTypeSymbol>(allInterfaces.Length + 1);
-            result.AddRange(allInterfaces);
-            result.Add(namedType);
-            return result;
+            if (SymbolEqualityComparer.Default.Equals(@interface, namedType))
+                return allInterfaces;
         }
 
-        return allInterfaces;
+        return allInterfaces.Add(namedType);
     }
 
     public static IEnumerable<ISymbol> GetAllMembers(this ITypeSymbol? symbol)
--- a/tests/Meziantou.Framework.Roslyn.Tests/RoslynHelperTests.cs
+++ b/tests/Meziantou.Framework.Roslyn.Tests/RoslynHelperTests.cs
@@ -764,7 +764,37 @@ public sealed class RoslynHelperTests
             """);
         var type = GetRequiredType(compilation, "ISample");
 
-        Assert.Contains(type, type.GetAllInterfacesIncludingSelf());
+        Assert.Contains(type, type.GetAllInterfacesIncludingSelf(), SymbolEqualityComparer.Default);
+    }
+
+    [Fact]
+    public void GetAllInterfacesIncludingThis_DoesNotIncludeTheTypeWhenSymbolIsNotAnInterface()
+    {
+        var compilation = CreateCompilation("""
+            public interface ISample;
+            public class Sample : ISample;
+            """);
+        var type = GetRequiredType(compilation, "Sample");
+        var interfaceType = GetRequiredType(compilation, "ISample");
+
+        var interfaces = type.GetAllInterfacesIncludingSelf();
+
+        Assert.Contains(interfaceType, interfaces, SymbolEqualityComparer.Default);
+        Assert.DoesNotContain(type, interfaces, SymbolEqualityComparer.Default);
+    }
+
+    [Fact]
+    public void GetAllInterfacesIncludingThis_DoesNotDuplicateSelfWhenAlreadyPresent()
+    {
+        var compilation = CreateCompilation("""
+            public interface IBase;
+            public interface ISample : IBase;
+            """);
+        var type = GetRequiredType(compilation, "ISample");
+
+        var interfaces = type.GetAllInterfacesIncludingSelf();
+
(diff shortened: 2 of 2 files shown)


meziantou:
## CI status: the remaining failure is pre-existing and unrelated to this change

`build_and_test (windows-2025-vs2026, Debug, 2)` is red, but not because of anything in this PR.

**Every failing test across all failing jobs is in `Meziantou.Framework.TemporaryContainers.Tests`.** Nothing outside that project failed on any job, on either attempt.

### First attempt

Three jobs failed:

| Job | Failures | Cause |
|---|---|---|
| `ubuntu-26.04-arm, Debug, 2` | 1 / 13541 | `MongoDbContainerTests` — container refused connections (`SocketException (111)`) |
| `ubuntu-26.04, Debug, InvariantGlobalization, 2` | 1 / 13501 | same MongoDB timeout |
| `windows-2025-vs2026, Debug, 2` | 18 / 14065 | `DockerContainerTests` — Docker daemon unreachable |

The rest of the matrix — macOS, Windows, Linux x64 and arm, Debug and Release, invariant-globalization variants — passed.

### After re-running the failed jobs

Both Linux jobs went green, confirming the MongoDB failures were flakes. The Windows job failed identically, so it is deterministic rather than flaky.

### Root cause of the Windows failure

`ExecutableContainerRuntime.IsSupported()` treats a runtime as available when the CLI binary is found on `PATH`, without checking that the daemon is reachable. The `windows-2025` image ships `docker.exe` but runs no daemon, and no workflow starts one. The skip gate in `ContainerRuntimeTestsBase` depends on `IsSupported()`, so it never fires and the tests run against a dead daemon — 200 `npipe:////./pipe/docker_engine` errors.

Filed separately as #2041, since it is a public-API question in an unrelated component and does not belong in this PR.

### Why it surfaced here

`src/Meziantou.Framework.Roslyn/TypeSymbolExtensions.cs` is classified as a *global file*, so this change expanded the build to 159 projects rather than the usual handful. None of the last 30 runs' Windows jobs executed `DockerContainerTests` at all — this latent failure has simply not been reached in a while.

### This change itself

All 8 Roslyn test assemblies passed in every failing job (107/107 each, both target frameworks), and the full suite passes locally across Roslyn 4.8, 5.0, 5.3 and 5.6 — 214/214 per variant, on net10.0 and net11.0, built with `/p:TreatsWarningsAsErrors=true`.

PR #2047, opened 2026-08-27
Ship the highlight.js third-party notice in the SyntaxHighlighting package

meziantou:
## Problem

`Meziantou.Framework.SyntaxHighlighting` is derived from [highlight.js](https://highlightjs.org/) v11.10.0 (BSD-3-Clause), and the repository correctly maintains `THIRD_PARTY_NOTICES.md` — but **that file never reaches consumers**. Packing v1.0.1 produces:

```
LICENSE.txt
README.md
icon.png
lib/net10.0/Meziantou.Framework.SyntaxHighlighting.dll
lib/net11.0/Meziantou.Framework.SyntaxHighlighting.dll
```

BSD-3-Clause requires binary redistributions to reproduce the copyright notice and license text "in the documentation and/or other materials provided with the distribution". A NuGet package is a binary redistribution, so the notice needs to be in it.

Two smaller problems in the same file: it points at `Highlighter/Engine/` and `Highlighter/Languages/`, but the actual directories are `Engine/` and `Languages/`.

## Changes

- Pack `THIRD_PARTY_NOTICES.md` into the package root.
- Correct the two stale directory paths inside the notice.
- Add a short "Third-party code" section to `readme.md` so the attribution is visible on nuget.org.

No code changes.

## Verification

Packed the project and listed the result — the notice is now included:

```
LICENSE.txt
README.md
THIRD_PARTY_NOTICES.md          <-- added
lib/net10.0/Meziantou.Framework.SyntaxHighlighting.dll
lib/net11.0/Meziantou.Framework.SyntaxHighlighting.dll
```

`dotnet run ./eng/update-all.cs` produces no further changes. Build is clean with `-p:TreatWarningsAsErrors=true`; all 4237 tests pass.

## Note

Other ported projects in this repository may have the same gap — worth a sweep.

---
Found during a review of `src/Meziantou.Framework.SyntaxHighlighting`.

### DIFF
--- a/src/Meziantou.Framework.SyntaxHighlighting/Meziantou.Framework.SyntaxHighlighting.csproj
+++ b/src/Meziantou.Framework.SyntaxHighlighting/Meziantou.Framework.SyntaxHighlighting.csproj
@@ -6,4 +6,9 @@
     <Version>1.0.1</Version>
   </PropertyGroup>
 
+  <ItemGroup>
+    <!-- BSD-3-Clause requires binary redistributions to reproduce the highlight.js notice -->
+    <None Include="THIRD_PARTY_NOTICES.md" Pack="true" PackagePath="" />
+  </ItemGroup>
+
 </Project>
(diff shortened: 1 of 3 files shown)


meziantou:
The fix must be in  https://github.com/meziantou/Meziantou.NET.Sdk/blob/1a0bff8bee699d90827f09e30d144ec2fb59cc8c/src/common/Common.targets#L160

PR #2049, opened 2026-08-27
Re-pin the output buffer when retrying FSCTL_READ_FILE_USN_DATA

meziantou:
## The defect

`ChangeJournal.GetEntry(SafeFileHandle)` retried `DeviceIoControl` with a stale pointer and a larger length, handing the kernel permission to write past the end of a pinned managed array.

```csharp
var buffer = new byte[USN_RECORD_V3.SizeOf(512)];
fixed (void* bufferPtr = buffer)          // pins the 512-byte-name allocation
{
    ...
    if (errorCode == (int)WIN32_ERROR.ERROR_MORE_DATA)
    {
        buffer = new byte[returnedSize];  // new, larger array — bufferPtr still points at the OLD one
        controlResult = PInvoke.DeviceIoControl(..., bufferPtr, (uint)buffer.Length, ...);
        //                                           ^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^
        //                                           old address, new larger size
    }
```

`bufferPtr` is bound by the enclosing `fixed` to the *original* allocation and never re-pinned. The retry tells the driver it may write `buffer.Length` bytes — the new, larger size — at an address backed by the smaller array. That is an out-of-bounds kernel write into the managed heap.

`Natives/Win32DeviceControl.cs:32-36` performs the same grow-and-retry correctly by opening a fresh `fixed` block over the new array, which is what marks this as a slip rather than an intended pattern.

**Reachability is narrow.** `FSCTL_READ_FILE_USN_DATA` returns a V2/V3 record, and NTFS caps a file name at 255 UTF-16 code units (510 bytes), so the 512-byte allowance normally suffices and the branch is not taken. The branch is unconditionally wrong when it *is* taken, and heap corruption from a kernel write is about the worst thing to have to debug — the crash surfaces far from the cause, in unrelated managed objects.

## What changed

- The call moved into a loop, with the `fixed` block *inside* it, so every attempt pins the buffer it actually passes.
- The `512` magic number became `MaximumFileNameLengthInBytes = 255 * sizeof(char)`, naming the NTFS limit the value was silently derived from.
- Growth is now geometric and bounded by `MaximumEntryBufferLength` (64 KB). `ERROR_MORE_DATA` reports the bytes *written*, not the bytes *required*, so `new byte[returnedSize]` was not guaranteed to be big enough — the old code could have retried at the same insufficient size. The bound stops a driver that keeps reporting `ERROR_MORE_DATA` from growing the allocation without limit.
- Non-`ERROR_MORE_DATA` failures still throw `Win32Exception` with the original error code, unchanged.

`SafeHandleValue` now wraps the whole loop rather than being re-created per attempt, which keeps the handle referenced for the duration.

## Verification

Compiles clean for `net10.0` and `net11.0` with `-p:TreatWarningsAsErrors=true`.

**Not executed against a real NTFS volume** — written on macOS, where every journal test skips. The corrected retry path is also the one that was already effectively unreachable, so exercising it needs a file name longer than the buffer, which NTFS will not produce. The value here is that the code is no longer wrong if the branch is ever entered.

### DIFF
--- a/src/Meziantou.Framework.Win32.ChangeJournal/ChangeJournal.cs
+++ b/src/Meziantou.Framework.Win32.ChangeJournal/ChangeJournal.cs
@@ -17,6 +17,12 @@ namespace Meziantou.Framework.Win32;
 [SupportedOSPlatform("windows5.1.2600")]
 public sealed class ChangeJournal : IDisposable
 {
+    /// <summary>The NTFS file name limit is 255 UTF-16 code units, so a record's name never needs more than this.</summary>
+    private const int MaximumFileNameLengthInBytes = 255 * sizeof(char);
+
+    /// <summary>Upper bound on the buffer used to read a single USN record, so a driver that keeps reporting ERROR_MORE_DATA cannot grow it without limit.</summary>
+    private const int MaximumEntryBufferLength = 64 * 1024;
+
     private readonly bool _unprivileged;
 
     internal SafeFileHandle ChangeJournalHandle { get; }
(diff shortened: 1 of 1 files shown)


meziantou:
Correcting two things in this PR after a question about whether Windows long path support invalidates the 255 basis.

**The 255 is still right, but for a reason worth stating.** Long path support (Windows 10 1607+, `LongPathsEnabled` plus a `longPathAware` manifest) raises the *total path* limit from `MAX_PATH` 260 to roughly 32,767 characters. It does not change the *per-component* limit, which is [`lpMaximumComponentLength` from `GetVolumeInformation`](https://learn.microsoft.com/windows/win32/fileio/maximum-file-path-limitation) — documented as "commonly 255" and 255 on NTFS. That is the relevant limit here because a USN record's `FileName` holds [only the name of the file or directory](https://learn.microsoft.com/windows/win32/api/winioctl/ns-winioctl-usn_record_v3#members), not a path; reconstructing a path means walking `ParentFileReferenceNumber`. So a 32,000-character path still produces a record whose name is at most one component.

**But the constant had the wrong unit.** `USN_RECORD_V3.SizeOf(int count)` counts `char` elements, not bytes:

```csharp
int v = sizeof(USN_RECORD_V3);
if (count > 1) v += checked((count - 1) * sizeof(char));
```

I was passing `255 * sizeof(char)` = 510 into a parameter that means "number of characters", so the buffer was sized for a 510-character name — twice what is needed, and named `...InBytes` while being used as a count. Harmless at runtime, but the code claimed precision it did not have, which is worse than the `512` magic number it replaced. Now `MaximumFileNameLength = 255`, passed as a character count.

**Also raised the growth cap** from 64 KB to 128 KB. `FileNameLength` is a `ushort`, so a name can be up to 65,535 bytes, and the full record is that plus the struct — marginally over the old 64 KB ceiling. Unreachable in practice, but the cap should sit above the largest record the field can describe, not just below it.

Microsoft's own guidance ([Walking a Buffer of Change Journal Records](https://learn.microsoft.com/windows/win32/fileio/walking-a-buffer-of-change-journal-records)) is to size from `MaxComponentLength` rather than hardcode it. I kept the constant and documented it as a starting size, since the retry loop this PR fixes is what actually handles a volume whose components exceed it — querying `GetVolumeInformation` per volume would be the thorough version if you'd prefer it.

Rebuilt clean for net10.0 and net11.0 with `-p:TreatWarningsAsErrors=true`.

PR #2108, opened 2026-08-27
Stop rewriting ASCII characters in ReplaceConfusablesCharacters

meziantou:
**Stack 1 of 2 · merges into `main` · must merge before #2109**

The confusables table contains eight sources below U+0080, and they were applied
unconditionally:

| source | replacement |
|---|---|
| `"` | `''` |
| `%` | `º/₀` |
| `0` | `O` |
| `1` | `l` |
| `I` | `l` |
| `` ` `` | `'` |
| `m` | `rn` |
| `\|` | `l` |

So `Unicode.ReplaceConfusablesCharacters("Item 1 of 10")` returned **`"ltern l of lO"`**.

The XML doc said the method returns "The text with confusable characters replaced" and the
readme example ends with `Console.WriteLine(normalized)`, both presenting the result as
displayable text. A caller following the readme to normalize a username, filename or display
name before **storing** it corrupted nearly every realistic input, silently and irreversibly.

The reason this stayed hidden is that the readme's own example string, `"paypal"`, happens to
contain none of the eight.

## What changed

ASCII is the alphabet this table normalizes *toward*, so an ASCII source is already in its
canonical form and folding it cannot make anything less confusable. `LoadConfusablesEntries`
now skips sources below U+0080, and the generated table drops from 6,565 to 6,557 entries.

`Unicode.ReplaceConfusablesCharacters("раураl")` still returns `"paypal"` — every Cyrillic,
Greek and other non-ASCII look-alike is untouched by this change.

Docs are corrected to match:

- The `<seealso href="https://unicode.org/reports/tr39/" />` references are removed. This was
  never the `skeleton` algorithm from UTS #39 — that is defined as
  `toNFD(toConfusable(toNFD(X)))` applied to a fixed point, while this is a single pass over
  un-normalized input — and with ASCII sources excluded it is further from it still. Citing
  the spec invited callers to use this for homograph detection, which it does not do
  correctly.
- The XML docs and readme now say the result is displayable text, not a comparison key, and
  say so explicitly rather than by omission.
- The generated file's `Purpose:` header line, which claimed the table "helps detect visually
  confusable characters that may be used in security attacks", is updated.

A proper UTS #39 skeleton API for the security use case is worth adding separately; this PR
deliberately does not pretend the existing method is one.

`<Version>` is intentionally left at 2.0.1 — the generator bumps it automatically on any data
change, and that bump is reverted here since version bumps go through the bot's own PRs.

## Verification

`dotnet test ... /p:TreatsWarningsAsErrors=true` — **44 passed** (22 facts x 2 TFMs), up from
32, 0 failed.

New facts: four ASCII strings returned by reference (`Assert.Same`, which also pins the
no-allocation fast path), the Cyrillic `paypal` fold still working, and
`IsConfusableCharacter` returning false across the printable ASCII range.

The generator was run against live unicode.org data and reports
`6557 confusable mappings ... 8 ASCII sources skipped`; the generated diff is exactly those
eight entries plus the header statistics.

Found by a project review of `src/Meziantou.Framework.Unicode`.

### DIFF
--- a/src/Meziantou.Framework.Unicode/readme.md
+++ b/src/Meziantou.Framework.Unicode/readme.md
@@ -1,6 +1,6 @@
 # Meziantou.Framework.Unicode
 
-This package provides Unicode helpers for normalizing confusable characters using the Unicode confusables table.
+This package provides Unicode helpers for normalizing characters that look like other characters, using the Unicode confusables table.
 
 ```csharp
 using Meziantou.Framework;
@@ -10,6 +10,17 @@ var normalized = Unicode.ReplaceConfusablesCharacters(input);
 
 Console.WriteLine(normalized); // "paypal"
 ```
+
+Characters that are already ASCII are never replaced, so ordinary text passes through untouched:
+
+```csharp
+Unicode.ReplaceConfusablesCharacters("Item 1 of 10"); // "Item 1 of 10"
+```
+
+> The result is displayable text, not a comparison key. This is deliberately **not** the
+> `skeleton` algorithm from [UTS #39](https://unicode.org/reports/tr39/): the input is not
+> normalized and the mapping is applied in a single pass, so it is not sufficient on its own
+> to decide whether two strings are confusable.
 
 This package also exposes Unicode character metadata from the Unicode data table:
 
--- a/src/Meziantou.Framework.Unicode/Unicode.Confusables.g.cs
+++ b/src/Meziantou.Framework.Unicode/Unicode.Confusables.g.cs
@@ -3,12 +3,13 @@
 // Data source: https://www.unicode.org/Public/UCD/latest/security/confusables.txt
 // Last modified: 2025-08-16T00:45:12.0000000+00:00
 // Statistics:
-//   - Total confusable mappings: 6565
-//   - Single-character replacements: 4462
-//   - Multi-character replacements: 2103
+//   - Total confusable mappings: 6557
+//   - Single-character replacements: 4457
+//   - Multi-character replacements: 2100
 //
 // Specification: https://www.unicode.org/reports/tr39/
-// Purpose: Helps detect visually confusable characters that may be used in security attacks
+// Purpose: Maps non-ASCII characters to the ASCII-ish character they resemble.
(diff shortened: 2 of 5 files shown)


src/Meziantou.Framework.Unicode/readme.md
```diff
@@ -11,6 +11,17 @@ var normalized = Unicode.ReplaceConfusablesCharacters(input);
 Console.WriteLine(normalized); // "paypal"
 ```
 
+Characters that are already ASCII are never replaced, so ordinary text passes through untouched:
+
+```csharp
+Unicode.ReplaceConfusablesCharacters("Item 1 of 10"); // "Item 1 of 10"
+```
+
+> The result is displayable text, not a comparison key. This is deliberately **not** the
```
meziantou, line 20: Implement the right algorithm

meziantou:
Closing in favour of implementing UTS #39 properly, per @meziantou's review comment.

Investigating that comment turned up a hard conflict with this PR: **the eight ASCII mappings
this PR deletes are load-bearing for the real algorithm.** Measured against the full table,
384 characters in U+0020..U+2FFF are unmapped themselves but decompose to something mapped —
`Ï` (U+00CF) has NFD `I` + U+0308, and `I` -> `l` is one of the eight. Remove them and
`skeleton()` stops folding `Ï`, `Ì`, `Í`, `Î`, so it is no longer conformant.

Filtering the table and implementing the standard are therefore mutually exclusive, and the
standard wins. The ASCII text-mangling this PR was fixing —
`ReplaceConfusablesCharacters("Item 1 of 10")` returning `"ltern l of lO"` — is real but is a
separate question about what that method is for, and will be decided on its own rather than by
quietly dropping data the algorithm needs.

Useful things learned here, carried into the replacement work:

- The confusables table **is** closed under its own mapping (0 of 6,565 targets contain a
  mapped character), so a conformant `skeleton()` needs a single mapping pass, not fixed-point
  iteration.
- The correct shape is NFD -> map -> NFD.

Superseded by the UTS #39 implementation. The allocation work that was stacked on top of this
(#2109) is independent of the table and is being reopened against `main` directly.

PR #2115, opened 2026-08-27
Distinguish an absent cgroup file from an unlimited value

meziantou:
## Finding

`ReadFile` mapped `FileNotFoundException` and `DirectoryNotFoundException` to `""`, and every getter mapped `""` — and unparseable content, and the literal `"max"` — to `null`. So `GetMemoryMax() == null` meant **any** of:

1. the memory controller is not enabled on this cgroup,
2. the cgroup does not exist at all,
3. the limit is genuinely unlimited,
4. the file contained something the parser did not understand.

A caller cannot write correct code against that. Deciding whether to apply a limit based on `GetMemoryMax() is null` treats "controller not enabled" as "unlimited", which is exactly backwards for the safety decision it is usually driving.

## Change

The getters that read a single interface file now return `CGroupValue<T>`, which carries the outcome alongside the value. The first revision of this PR threw for the absent cases; it now reports them, per review feedback.

| `State` | meaning |
|---|---|
| `Configured` | the file holds a value, available through `Value` |
| `NotConfigured` | the file holds no limit (`max`) |
| `Unavailable` | the file does not exist: the controller is not enabled on the parent cgroup, or the kernel does not support the feature |
| `Invalid` | the file holds content this library does not understand |

Four states rather than three: "no limit set" and "controller not enabled" are the two cases the finding is about, so collapsing them back into one `NotConfigured` would leave `GetMemoryMax()` unable to answer *"is this process actually capped?"*.

```csharp
var limit = group.GetMemoryMax();
var effective = limit switch
{
    { State: CGroupValueState.Configured } v => v.Value,
    { State: CGroupValueState.NotConfigured } => long.MaxValue,     // genuinely unlimited
    { State: CGroupValueState.Unavailable } => ApplyOwnLimit(),     // no memory controller — do not assume unlimited
    var v => throw new InvalidOperationException($"unexpected memory.max: {v.RawValue}"),
};
```

`RawValue` keeps the unparsed content, so a kernel format change is diagnosable and a caller can still read a value this library predates — strictly more information than throwing gave. `TryGetValue` and `GetValueOrDefault(fallback)` cover the callers that do not care why a value is missing; `Value` throws `InvalidOperationException` when there is none. `default(CGroupValue<T>)` is `Unavailable`.

Two getters deliberately keep their shape:

- **`IsFrozen()` stays `bool`.** The freezer needs Linux 5.2+, and a kernel without a freezer cannot have frozen the cgroup, so an absent file is a correct `false` rather than a missing value.
- **`GetCpuMax()` is never `NotConfigured`.** The period stays meaningful without a quota, so an unlimited quota is reported as `CpuMax.MaxMicroseconds` being `null` rather than by a state that would throw the period away.

The four `IEnumerable` getters (`GetProcesses`, `GetThreads`, `GetAvailableControllers`, `GetEnabledControllers`) also keep their shape — a deferred sequence and a state wrapper mix badly, and those files always exist in a live cgroup.

Reading a cgroup that no longer exists now throws `DirectoryNotFoundException` instead of reporting absent values, so a stale `CGroup2` no longer looks like a healthy cgroup without controllers. The readme already documented that exception.

## Drive-by fix

`GetAvailableControllers` and `GetEnabledControllers` now trim before splitting. Without it the last controller carried a trailing `\n`, so `GetAvailableControllers().Contains("misc")` was false whenever `misc` came last — which also silently affected `ProcessWrapper.EnsureControllerIsAvailable`.

## ⚠️ Breaking change

About 20 getter signatures change, so `ref/` changes and this needs a major version bump. `<Version>` is left alone here, since the bot bumps versions in its own PR.

Migration is mechanical:

```csharp
// before
if (group.GetMemoryMax() is long limit) { … }

// after
if (group.GetMemoryMax().TryGetValue(out var limit)) { … }
```

## Verification

`dotnet build -p:TreatWarningsAsErrors=true` clean for `net10.0` and `net11.0`. `ref/` regenerated and verified with `PublicApiGeneratorVerifyNoChangeOnBuild=true`. `dotnet run ./eng/update-all.cs` produced no changes.

The parsers moved into `internal static` methods taking the file content, so they are now covered on every OS rather than only on a privileged Linux host: 25 new tests in `CGroup2ParsingTests` (the four-outcome table, raw-value retention, `cpu.max` period preservation, `CGroupValue` semantics), plus the six existing integration tests updated. `148 total, 124 passed, 24 skipped (Linux-only), 0 failed`.


meziantou:
The idea is ok, but the execution is not. Throwing exceptions is not great. I think it should return a type (not configured, configured, invalid)

PR #2119, opened 2026-08-27
Skip resx entries whose generated member name is already taken

meziantou:
> **Stack 3 of 3** · merges into `fix/typed-inline-values` (#2118) · merge last.

## The bug

`ToCSharpNameIdentifier` maps every character outside the identifier categories to `_`, with no uniqueness check afterwards, and the emitted members were never checked against the fixed members the generator always declares. Three collisions, all verified by compiling the generated output:

- **Two names normalizing to one identifier.** `Hello World` and `Hello-World` both become `Hello_World`.
- **A name shadowing the boilerplate.** An entry called `ResourceManager`, `Culture`, `GetString`, `GetObject`, `GetStream` or `GetStringWithDefault` collides with a member of the generated class.
- **A name colliding with a generated `Format*` method.** An entry `Hello` with a placeholder generates `FormatHello`; an entry literally named `FormatHello` generates a property of the same name.

A resx with the first two produced:

```
error CS0102: The type 'test' already contains a definition for 'Hello_World'
error CS0102: The type 'testNames' already contains a definition for 'Hello_World'
error CS0102: The type 'test' already contains a definition for 'ResourceManager'
error CS0229: Ambiguity between 'test.ResourceManager' and 'test.ResourceManager'   (×3)
```

Spaces, hyphens and dots in resx keys are common in real localization files. The user sees `CS0102` pointing into a `.g.cs` they never wrote, and the `CS0229` cascade makes the generator's own boilerplate look broken, which buries the actual cause.

## What changed

A `HashSet<string>` per generated type, seeded with the names that type always declares — including the class name itself, since a member may not share its enclosing type's name (CS0542). Each entry reserves its member name before emitting; a name already taken is skipped with a comment saying why. `Format*` method names are reserved in the same set, so the third collision is covered too.

This is the same shape as the existing `GetHintName` collision loop at `ResxGenerator.cs:114-127`, which already solves this problem well for generated file names.

**Skip rather than rename**, deliberately: a silently renamed property is its own surprise, and the resx author almost certainly meant these to be two different resources. As in stack 2, this really wants a diagnostic instead of a comment — blocked on the generator not reporting diagnostics at all, which is a separate change.

## Verification

- `ResourceNamesMappingToTheSameIdentifierDoNotCollide` — `Hello World` + `Hello-World`.
- `ResourceNamesCannotShadowTheGeneratedMembers` — a theory over `ResourceManager`, `Culture`, `GetString`, `GetObject`, `GetStream`.
- `AFormatMethodDoesNotCollideWithAResourceOfTheSameName` — `Hello` (with a placeholder) + `FormatHello`.

The test helper asserts the generated code actually compiles, which is the real assertion here. All 12 collision cases fail against the unguarded loop. 66/66 unit tests and 16/16 generator tests pass on net10.0 and net11.0.


src/Meziantou.Framework.ResxSourceGenerator/ResxGenerator.cs
```diff
@@ -297,11 +297,32 @@ private static string GenerateCode(string? ns, string className, string? resourc
...
+
             foreach (var entry in entries.OrderBy(e => e.Name, StringComparer.Ordinal))
             {
                 if (string.IsNullOrEmpty(entry.Name))
                     continue;
 
+                var memberName = ToCSharpNameIdentifier(entry.Name);
+                if (!memberNames.Add(memberName))
+                {
+                    sb.AppendLine("        // Skipped '" + memberName + "': the name is already used by another member");
```
meziantou: I think it should generate a deterministic unique name instead of skipping the entry

meziantou: Done in f35362c5b — colliding entries now get a deterministic numeric suffix instead of being dropped.

`GetUniqueMemberName` mirrors the existing `GetHintName` loop: take the name if free, otherwise the first of `name1`, `name2`... that is. Entries are iterated in `OrderBy(e => e.Name, StringComparer.Ordinal)`, so the assignment is stable across builds and machines.

Two things this changed beyond the literal request:

- **Format methods are uniquified too**, rather than dropped. With entries `Hello` (`"Hello {0}!"`) and `FormatHello`, `FormatHello` sorts first and takes the plain name, so `Hello`'s format method becomes `FormatHello1`. Previously that method was silently not generated.
- **The key-names type now uses the same ordering.** It iterated `entries` in file order while the resource type iterated sorted by name, so with `Hello World` / `Hello-World` the two types could disagree about which entry `Hello_World` referred to — `Resources.Hello_World` and `ResourcesNames.Hello_World` pointing at different resx entries. Both loops now sort identically, and a test asserts the mapping (`@Hello_World = "Hello World"`, `@Hello_World1 = "Hello-World"`).

Tests assert member names off the generated syntax tree rather than by string matching, since the generator emits `Environment.NewLine` and a newline-anchored `Assert.Contains` would have failed on Windows. 86/86 unit and 16/16 generator tests pass on net10.0 and net11.0; all 14 collision cases fail if the uniquifier is neutralized.

PR #2122, opened 2026-08-27
Remove the AAA comments and poll instead of sleeping in the freeze test

meziantou:
**Stack 2 of 2 · merges into `test/fix-process-stranding` · merge after #2121**

## Findings

Two conventions issues in the same file.

**AAA comments.** `AGENTS.md` states: "When writing tests, do not emit 'Act', 'Arrange' or 'Assert' comments." `CGroup2Tests.cs` used them in essentially every test.

**A sleep-based assertion.** `Freeze_ShouldFreezeProcesses` slept 100 ms and then asserted `IsFrozen()` (`:250-254`). Freezing a cgroup is asynchronous, so on a loaded machine the assertion can run before the kernel has transitioned the cgroup, and the test fails spuriously. Because this suite only runs locally, that flake lands on a developer mid-task.

## Change

- Removed the bare `// Arrange` / `// Act` / `// Assert` / `// Cleanup` marker comments. The one descriptive comment that starts with "Cleanup:" is kept — it explains what `Dispose` does rather than labelling a section.
- Replaced the fixed sleep with a bounded poll: up to 5 s at 25 ms intervals. Faster in the common case and not load-sensitive.

No behavior change to what any test asserts.

## Verification

`dotnet build -p:TreatWarningsAsErrors=true` clean for `net10.0` and `net11.0`. Same caveat as #2121 — these tests are skipped outside a privileged Linux environment, so they were not executed.

## Why this is stacked

The AAA comment removal touches lines *inside* the test that #2121 rewrites. As independent PRs these would conflict, and folding them together would bury #2121's five-line correctness fix inside a forty-line comment deletion.

### DIFF
--- a/tests/Meziantou.Framework.Unix.ControlGroups.Tests/CGroup2Tests.cs
+++ b/tests/Meziantou.Framework.Unix.ControlGroups.Tests/CGroup2Tests.cs
@@ -210,7 +210,7 @@ public sealed class CGroup2Tests : IDisposable
     }
 
     [Fact]
-    public void Freeze_ShouldFreezeProcesses()
+    public async Task Freeze_ShouldFreezeProcesses()
     {
         if (!File.Exists(Path.Combine(_testRoot.Path, "cgroup.freeze")))
             throw new Exception("$XunitDynamicSkip$Freezer not available");
@@ -226,7 +226,7 @@ public sealed class CGroup2Tests : IDisposable
             isFrozen = child.IsFrozen();
             if (!isFrozen)
             {
-                Thread.Sleep(25);
+                await Task.Delay(25, XunitCancellationToken);
             }
         }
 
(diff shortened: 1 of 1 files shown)


tests/Meziantou.Framework.Unix.ControlGroups.Tests/CGroup2Tests.cs
```diff
@@ -250,18 +215,23 @@ public void Freeze_ShouldFreezeProcesses()
...
-        // Assert
-        var isFrozen = child.IsFrozen();
+        // Freezing is asynchronous, so poll instead of racing a fixed sleep against a loaded machine.
+        var isFrozen = false;
+        for (var i = 0; i < 200 && !isFrozen; i++)
+        {
+            isFrozen = child.IsFrozen();
+            if (!isFrozen)
+            {
+                Thread.Sleep(25);
```
meziantou: Use Task.Delay and make method async

PR #2129, opened 2026-08-27
Add tests for the cgroup file parsers

meziantou:
**Stack 1 of 2 · merges into `main` · must merge before #2130**

## Finding

**No test in this project has ever run in CI.** The `CGroup2Tests` fixture constructor skips the whole class unless the process is Linux, has cgroup v2 mounted, is privileged, **and** is not on GitHub Actions (`CGroup2Tests.cs:11-42`). Every CI test job runs on a hosted runner, so the GitHub Actions gate alone skipped all twelve tests on every run — and `MinimumExpectedTests=0` meant nothing complained.

Those skips are defensible: you genuinely cannot create cgroups on a hosted runner. The problem is what got concluded from them. The code most likely to be quietly wrong needs none of those privileges:

- `CpuStat.Parse`, `MemoryStat.Parse` — already `internal` and visible to this assembly
- `ParseCpuList`, `ConvertToRanges` — pure string functions

None had a test. Every parser regression shipped undetected.

## Change

New `CGroup2ParsingTests` — 21 tests over the four pure functions, using recorded fixture text in the real `cpu.stat` / `memory.stat` / `cpuset.cpus` formats. `ParseCpuList` and `ConvertToRanges` change from `private` to `internal` to make them reachable; no public API change, so `ref/` is untouched.

`MinimumExpectedTests` goes from `0` to `21`, so a future skip regression is caught rather than silently passing. The stale comment above it — which claimed *all* tests are skipped — is corrected.

Cases now covered that weren't: absent bandwidth keys in `cpu.stat`, unknown and malformed lines, reversed ranges (`"5-3"`), non-numeric input, empty and whitespace lists, unsorted input to `ConvertToRanges`, and a round trip through both directions.

Three tests deliberately assert current behavior that is arguably wrong, each labelled as documenting a known limitation rather than endorsing it:

- `memory.stat` values above `long.MaxValue` (kernel counters are `u64`) silently leave the field at `0` — #2130 changes this.
- `kernel` and `swapcached`, absent on older kernels, are indistinguishable from zero — #2130 changes this too.
- `ConvertToRanges([1, 1, 2])` yields `"1,1-2"` because the input is not de-duplicated.

## Verification

`dotnet build -p:TreatWarningsAsErrors=true` clean for `net10.0` and `net11.0`.

Full suite run, both target frameworks: **42 passed, 0 failed, 24 skipped** (21 new tests × 2 TFMs; the skips are the pre-existing privileged integration tests). These run on macOS, Windows and Linux, so they execute on every CI leg.

The privileged integration tests stay gated exactly as they were. Getting real cgroup coverage into CI would need a privileged container step — out of scope here.

### DIFF
--- a/tests/Meziantou.Framework.Unix.ControlGroups.Tests/Meziantou.Framework.Unix.ControlGroups.Tests.csproj
+++ b/tests/Meziantou.Framework.Unix.ControlGroups.Tests/Meziantou.Framework.Unix.ControlGroups.Tests.csproj
@@ -2,9 +2,6 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(LatestTargetFrameworks)</TargetFrameworks>
-    <!-- The CGroup2Tests integration tests are skipped when not running on Linux, when cgroup v2 is unavailable,
-         or when running on GitHub Actions. The parser tests in CGroup2ParsingTests need none of that and always run. -->
-    <MinimumExpectedTests>21</MinimumExpectedTests>
     <PredefinedCulturesOnly>false</PredefinedCulturesOnly>
     <RootNamespace>Meziantou.Framework.Tests</RootNamespace>
   </PropertyGroup>
--- a/tests/Meziantou.Framework.Unix.ControlGroups.Tests/Meziantou.Framework.Unix.ControlGroups.Tests.csproj
+++ b/tests/Meziantou.Framework.Unix.ControlGroups.Tests/Meziantou.Framework.Unix.ControlGroups.Tests.csproj
@@ -2,8 +2,9 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(LatestTargetFrameworks)</TargetFrameworks>
-    <!-- All the tests are skipped when not running on Linux, when cgroup v2 is unavailable, or when running on GitHub Actions -->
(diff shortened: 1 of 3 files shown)


tests/Meziantou.Framework.Unix.ControlGroups.Tests/Meziantou.Framework.Unix.ControlGroups.Tests.csproj
```diff
@@ -2,8 +2,9 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(LatestTargetFrameworks)</TargetFrameworks>
-    <!-- All the tests are skipped when not running on Linux, when cgroup v2 is unavailable, or when running on GitHub Actions -->
-    <MinimumExpectedTests>0</MinimumExpectedTests>
+    <!-- The CGroup2Tests integration tests are skipped when not running on Linux, when cgroup v2 is unavailable,
+         or when running on GitHub Actions. The parser tests in CGroup2ParsingTests need none of that and always run. -->
+    <MinimumExpectedTests>21</MinimumExpectedTests>
```
meziantou:
```suggestion
```

PR #2140, opened 2026-08-27
Cover the stream serializer instead of rewinding the stream

meziantou:
`StreamSnapshotSerializer` copies from the stream's *current* position:

```csharp
var ms = new MemoryStream();
stream.CopyTo(ms);
```

The natural calling pattern leaves the position at the end:

```csharp
var ms = new MemoryStream();
await something.WriteToAsync(ms);
Snapshot.Validate(ms, SnapshotType.Png);   // snapshots zero bytes
```

`CopyTo` then yields nothing and the snapshot is written as an empty file.
`SnapshotEngine.Validate` only rejects a zero-*element* `SerializedSnapshot`
(`SnapshotEngine.cs:19-20`), not a one-element snapshot holding zero bytes, so nothing
catches it. The first run writes an empty `.verified` file and every subsequent run passes
against it — a test that asserts nothing while appearing green.

## What changed

- Seek to the beginning when `stream.CanSeek`, leaving non-seekable streams (network,
  pipes, compression) read from wherever they are.
- Dispose the intermediate `MemoryStream`, which was previously leaked.

## Verification

Two tests added: one for a `MemoryStream` positioned at the end, one for a non-seekable
stream. The existing `DefaultSerializer_HandlesStreamAsBinary` still asserts the stream is
consumed to its end.

```
dotnet build slnx/Meziantou.Framework.SnapshotTesting.slnx /p:TreatsWarningsAsErrors=true   # 0 warnings, 0 errors
dotnet test tests/Meziantou.Framework.SnapshotTesting.Tests --filter-not-class "*EndToEnd*" # 116/116 passed
```

## Note

This changes the caller's stream position as a side effect. That matches what a snapshot
assertion is asking for (the whole content), but it is a behaviour change for anyone who
was relying on partial reads.


meziantou:
I don't think the change is ok. Rewinding may change the behavior.
However, I like having more tests about the serializer.

meziantou:
Fair enough — dropped the rewind in d667407. `StreamSnapshotSerializer` reads from the current position exactly as before; the only remaining code change is disposing the intermediate `MemoryStream`, which was previously leaked. Say the word if you want that out too and I will leave the file untouched.

Kept and reshaped the tests so they pin the behaviour as it actually is rather than as I had proposed to change it:

- `DefaultSerializer_HandlesStreamAsBinary` — unchanged, pre-existing.
- `DefaultSerializer_ReadsTheStreamFromItsCurrentPosition` — a stream positioned mid-way snapshots only the remainder. This is the case your objection protects, so it is now covered.
- `DefaultSerializer_ProducesAnEmptySnapshotForAStreamPositionedAtTheEnd` — a caller that has just written to a `MemoryStream` snapshots zero bytes. That was the trap I originally wanted to fix; pinning it means the behaviour is at least deliberate and visible rather than incidental.
- `DefaultSerializer_HandlesNonSeekableStream` — non-seekable streams work.

Also retitled the PR, since "rewind" no longer describes it.

One thing worth naming, purely so it is a known trade-off rather than a surprise: the empty-snapshot case fails quietly. `SnapshotEngine.Validate` rejects a zero-*element* `SerializedSnapshot` but not a one-element snapshot holding zero bytes, so the first run writes an empty `.verified` file and every later run passes against it — a test that asserts nothing while looking green. Not proposing anything here; if you ever want it addressed, a serializer-level guard or an engine-level "empty snapshot needs opt-in" would both work without touching the read-from-current-position contract.

PR #2146, opened 2026-08-27
Enable Exited before starting the git merge tool

meziantou:
`GitMergeTool` copies the verified file aside, launches the merge tool, and subscribes to
`Exited` to delete the copy afterwards:

```csharp
var process = Process.Start(filename, args);
process.Exited += (sender, args) => { /* delete the temp copy; process.Dispose() */ };
```

`Process.EnableRaisingEvents` defaults to `false`, so `Exited` never fires. The handler is
dead code: the copy is never deleted and the `Process` is never disposed.

The leak is systemic rather than local. `MergeTool.CopyFileToTemp` created a fresh
GUID-named directory directly under the temp path on every call, and
`VisualStudioMergeTool` (`VisualStudioMergeTool.cs:18`) calls it with no cleanup attempt at
all. Every failing snapshot assertion that opens a merge tool left one directory plus one
file behind permanently — and a developer iterating on snapshots does that dozens of times
a day.

## What changed

- `GitMergeTool` sets `EnableRaisingEvents = true` before subscribing, so the existing
  cleanup handler actually runs.
- `CopyFileToTemp` now creates its GUID directories under a single
  `%TEMP%/Meziantou.Framework.SnapshotTesting/` root, tracks them, and deletes them on
  `ProcessExit`. That covers `VisualStudioMergeTool` and any future tool without each one
  re-implementing cleanup. The GUID subdirectory is kept so the copy keeps its original
  file name, which is what the diff tool shows in its header.

## Verification

Added `CopyFileToTemp_PlacesCopiesUnderTheSharedRootAndCleansThemUp`, covering the
location, the preserved file name, the content, and that cleanup removes the directory.

```
dotnet build slnx/Meziantou.Framework.SnapshotTesting.slnx /p:TreatsWarningsAsErrors=true   # 0 warnings, 0 errors
dotnet test tests/Meziantou.Framework.SnapshotTesting.Tests --filter-not-class "*EndToEnd*" # 115/115 passed
```

The `EnableRaisingEvents` line itself is not covered by a test — exercising it needs a real
merge tool process. Cleanup on `ProcessExit` is also best-effort: it will not run if the
test host is killed, which is why the copies now share one root that can be cleared
wholesale.


src/Meziantou.Framework.SnapshotTesting/MergeTools/GitMergeTool.cs
```diff
@@ -22,6 +22,10 @@ internal sealed class GitMergeTool : GitTool
                          .Replace("$MERGED", currentFilePath, StringComparison.Ordinal));
 
                 var process = Process.Start(filename, args);
+
+                // Exited never fires unless the process is asked to raise it, and without it the copy below
+                // is never deleted and the Process handle is never released.
+                process.EnableRaisingEvents = true;
```
meziantou: Should it be set before Process.Start?

src/Meziantou.Framework.SnapshotTesting/MergeTool.cs
```diff
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
```
meziantou: Revert this file. The temp files can live longer than the process

src/Meziantou.Framework.SnapshotTesting/MergeTools/GitMergeTool.cs
meziantou: Yes — moved it before `Start` in ff7db11.

There is a real race the other way round: `Process.Start(filename, args)` returns an already-running process, so a merge tool that exits immediately could be gone before `EnableRaisingEvents = true` and the subscription land, and the temp copy would leak exactly as it did before. Setting it up front removes that window.

That meant switching from the static `Process.Start` factory to constructing the `Process`, which surfaced a genuine leak the factory had been hiding: if `Start()` throws, the instance is never disposed. Added a try/catch for that. CA2000 still flags the ownership transfer to `ProcessMergeToolResult`, so I suppressed it with the same justification `SystemProcessFactory.cs` uses for the identical pattern.

src/Meziantou.Framework.SnapshotTesting/MergeTool.cs
meziantou: You are right, and the reason is worse than untidiness — reverted in ff7db11.

`MergeToolStrategy` launches the tool and returns immediately, so the diff tool routinely outlives the test process. Deleting the temp copies on `ProcessExit` would have pulled files out from under a tool the developer still had open. That is a regression, not a cleanup, and I missed it.

`MergeTool.cs` is now byte-identical to main. What remains in the PR is only the `GitMergeTool` fix, which ties each copy's lifetime to the process that actually uses it via `Exited` — the correct lifetime, and the one that was silently never firing.

Also dropped `CopyFileToTemp_PlacesCopiesUnderTheSharedRootAndCleansThemUp`, since it tested the reverted API.

For the record, so it is not lost: `VisualStudioMergeTool` (`VisualStudioMergeTool.cs:18`) also calls `CopyFileToTemp` and never cleans up, so those copies still accumulate. Any fix there has the same constraint you identified — the lifetime has to follow the tool process, not the test process. Happy to open an issue if you want it tracked.

PR #2164, opened 2026-08-27
Make the HAR model nullable so archives with an explicit null still parse

meziantou:
Stack 1 of 2 · merges into `main` · must merge before #2165

> Updated after review: the first version made the parser reject an explicit `null`. Per feedback, it now makes the model nullable instead, so more documents parse. Rebased onto current `main` and squashed to a single commit; #2165 is rebased on top of it.

## The problem

`System.Text.Json` writes `null` into a non-nullable reference property without complaint, so the model's annotations were not enforced by anything. Verified against the library before the fix:

- `{"log":null}` → `HarDocument.Parse` **succeeds**, `doc.Log` is `null`, and `doc.Log.Entries` throws `NullReferenceException`
- `{"log":{"entries":[{"request":{"method":null,"url":null,"headers":null},"response":null}]}}` → parses fine; `Request.Method`, `Request.Headers` and `Response` are all `null`, and `ToHttpRequestMessage()` throws `ArgumentNullException`

Callers were told by the type system that `document.Log.Entries` needed no null check. An archive from a buggy producer turned that into a `NullReferenceException` deep in user code with no indication the input was at fault.

## What changed

Every reference-typed property on the model is now nullable, so the annotations describe what the parser actually produces. Nothing is rejected: an archive that writes `null` anywhere still parses, and the possible `null` shows up at compile time rather than as a `NullReferenceException` at run time.

Unaffected:
- **Missing** properties still keep their initializer defaults (`Method` stays `""`, `Headers` stays an empty list). Only an explicit `null` yields `null`.
- On write, `DefaultIgnoreCondition.WhenWritingNull` omits those properties, so a document built in code serializes exactly as before.
- A top-level `null` document still surfaces as the existing `InvalidOperationException`.

The conversion helpers gained the matching null handling:
- `ToHttpRequestMessage()` / `ToHttpResponseMessage()` on `HarEntry` throw `InvalidOperationException` when the entry carries no request or response, and `ToHttpRequestMessage()` on `HarRequest` does the same when there is no method or URL — those cannot be reconstructed.
- A `null` (or empty) `mimeType` now falls back to the `StringContent` default instead of throwing out of `MediaTypeHeaderValue`.
- `null` header lists and headers without a name are skipped.

`HarHttpRecordingStore` reads through the nullable model and builds its documents up front rather than mutating the initializer defaults.

## Tradeoff

This is the alternative that the first version of this PR considered and did not take. It costs API churn — `ref/Meziantou.Framework.HttpArchive.cs` shows 36 properties changing shape, and callers that walk the graph now need `?.` or `!`, which the repo's own tests and readme show. In exchange no input is ever rejected, which is the priority for a format whose producers do not reliably follow the schema.

## Verification

`dotnet test -p:TreatsWarningsAsErrors=true` → **150/150 green** on `Meziantou.Framework.HttpArchive.Tests` and **76/76** on `Meziantou.Framework.Http.Recording.Tests`, across `net10.0` and `net11.0`. 13 new tests cover the null shapes that now parse, the missing-property defaults, the write path, and the conversion helpers' new exceptions. `dotnet run ./eng/update-all.cs` reports no further changes; `ref/` is regenerated and committed.

### DIFF
--- a/ref/Meziantou.Framework.HttpArchive.cs
+++ b/ref/Meziantou.Framework.HttpArchive.cs
@@ -7,9 +7,9 @@ namespace Meziantou.Framework.HttpArchive
     public sealed class HarBrowser
     {
         [System.Text.Json.Serialization.JsonPropertyName("name")]
-        public string Name { get => throw null; set { } }
+        public string? Name { get => throw null; set { } }
         [System.Text.Json.Serialization.JsonPropertyName("version")]
-        public string Version { get => throw null; set { } }
+        public string? Version { get => throw null; set { } }
         [System.Text.Json.Serialization.JsonPropertyName("comment")]
         public string? Comment { get => throw null; set { } }
         [System.Text.Json.Serialization.JsonExtensionData]
@@ -35,7 +35,7 @@ namespace Meziantou.Framework.HttpArchive
         [System.Text.Json.Serialization.JsonPropertyName("lastAccess")]
         public System.DateTimeOffset LastAccess { get => throw null; set { } }
         [System.Text.Json.Serialization.JsonPropertyName("eTag")]
-        public string ETag { get => throw null; set { } }
+        public string? ETag { get => throw null; set { } }
         [System.Text.Json.Serialization.JsonPropertyName("hitCount")]
         public int HitCount { get => throw null; set { } }
         [System.Text.Json.Serialization.JsonPropertyName("comment")]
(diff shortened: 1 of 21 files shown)


meziantou:
I would prefer the updating properties to be nullable, so we can parse more documents

meziantou:
Done — switched to nullable properties instead of rejecting the input. Every reference-typed property on the model is now nullable, so an archive with an explicit `null` anywhere parses. Missing properties still keep their initializer defaults (`Method` stays `""`, `Headers` stays an empty list), and `WhenWritingNull` omits nulls on write, so documents built in code serialize unchanged.

The conversion helpers picked up the matching handling: `ToHttpRequestMessage()`/`ToHttpResponseMessage()` throw `InvalidOperationException` when the entry has no request/response or the request no method/URL, a null `mimeType` falls back to the `StringContent` default instead of throwing out of `MediaTypeHeaderValue`, and unnamed headers are skipped.

Branch rewritten to a single commit, so #2165 needs a rebase.

PR #2167, opened 2026-08-27
Say plainly that diagnostics keep the caller's order

meziantou:
The comment on `DiagnosticSnapshotSerializer.TrySerialize` reads:

> The order is the caller's: unlike the generators of a driver run, an assertion on a list
> of diagnostics is usually about the order they were reported in. **Sort them before
> validating when the source of the list does not guarantee one.**

The subject shifts between the two sentences. The first describes the serializer, the
second addresses the *caller* — but read straight through, the last sentence looks like it
describes something the method does, and the loop below plainly does not sort. The sibling
`GeneratedSourcesSnapshotSerializer.cs:35-37` *does* sort its diagnostics, which makes the
apparent inconsistency look deliberate in the wrong direction.

I flagged this during review as a code/comment mismatch and initially planned to add the
sort for consistency. That would have been wrong:

- The caller's order is usually the thing under test — a source generator reporting
  diagnostics in a particular sequence is exactly what the snapshot is pinning.
- Sorting would silently rewrite every existing diagnostic snapshot for every consumer.

`GeneratedSourcesSnapshotSerializer` sorts because *its* input order genuinely is not a
contract (generators run in whatever order the driver picked), which its own comment says.
The two serializers are correct to differ.

## What changed

Reworded the comment so it cannot be read as describing the implementation, and states
explicitly that the serializer does not reorder and that sorting is the caller's job. No
behaviour change.

## Verification

Added `DiagnosticCollection_IsWrittenInTheOrderGiven`, which pins the ordering with
deliberately non-alphabetical messages, so a future "consistency" change like the one I
almost made fails a test instead of silently invalidating snapshots.

```
dotnet build slnx/Meziantou.Framework.SnapshotTesting.Roslyn.slnx /p:TreatsWarningsAsErrors=true   # 0 warnings, 0 errors
dotnet test tests/Meziantou.Framework.SnapshotTesting.Roslyn.Tests                                 # 20/20 passed
```


src/Meziantou.Framework.SnapshotTesting.Roslyn/DiagnosticSnapshotSerializer.cs
```diff
@@ -8,9 +8,10 @@ internal sealed class DiagnosticSnapshotSerializer : ISnapshotSerializer
 
     public bool TrySerialize(SnapshotType type, object? value, [NotNullWhen(true)] out SerializedSnapshot? result)
     {
-        // The order is the caller's: unlike the generators of a driver run, an assertion on a list of
-        // diagnostics is usually about the order they were reported in. Sort them before validating
-        // when the source of the list does not guarantee one.
+        // Diagnostics are written in the order they are given. Unlike the generators of a driver run, whose
```
meziantou: I think the original implementation is wrong. We should order the diagnostic, so output is deterministic

meziantou: Agreed and done in 269a916.

Diagnostics are now sorted before being written. Two things I found while implementing that are worth flagging:

**1. The sort key had to be culture-independent.** The obvious move was to copy `GeneratedSourcesSnapshotSerializer`, which sorted by `Id` then `Diagnostic.ToString()`. But `ToString()` formats the message with the current UI culture — the exact reason `RoslynFormatter.AppendDiagnostic` exists and doesn't use it. So that sort key was itself machine-dependent, and copying it would not have made the output deterministic.

Added `RoslynFormatter.DiagnosticComparer`, ordering by path → start position → id → invariant message, and used it in **both** serializers. That also removes the latent culture dependence from the generated-sources report. Shout if you would rather I left that one alone.

Position sorts before id so the report reads in the order a compiler would emit it.

**2. There was a pre-existing test pinning the old behaviour**, `Serialize_DiagnosticCollectionKeepsTheCallerOrder`, asserting the caller order was preserved. So the old behaviour was deliberate and covered, not accidental — worth knowing since this is a behaviour change that will require re-approving any existing diagnostic snapshot whose collection was not already in sorted order. I have updated it to `Serialize_DiagnosticCollectionIsSorted`.

Tests: replaced the ordering test with three — sorted output, order independent of input order, and position winning over id. 22/22 pass, Release build clean.

PR #2184, opened 2026-08-27
Drop the redundant jpg comparer registration and document the format sets

meziantou:
> **Stack 2 of 2 · merges into `fix/comparer-by-extension` (#2180) · merge that one first.**

Follow-up to #2180, which made comparers resolve by the format a snapshot is *stored* in.
This tidies up the registration helpers now that that holds.

## The redundant jpg registration

`AddImageComparer` ended with:

```csharp
comparers.Set(SnapshotType.Jpeg, comparer);
comparers.Set(SnapshotType.Create("jpg"), comparer);   // same key
```

`SnapshotType.Create("jpg")` resolves through the type cache to the `Jpeg` singleton, and
snapshot types compare by their canonical name — so the second call re-registered the key
the first had just set. Removed.

The existing `AddImageComparer_RegistersJpgAlias` and `AddImageComparer_RegistersTifAlias`
tests both still pass without it, which is the proof: `tif` never had an explicit
registration either, and it worked, because aliasing happens in `SnapshotType.Create`.

## Documenting what each helper covers

My review flagged that the three `Add*` helpers register three different, non-overlapping
sets of types and called that an inconsistency. Checking each backend, that framing was
wrong: **a comparer usefully reads formats its serializer cannot write.** `AddSkiaSharp`
registers BMP, GIF and ICO because `SKCodec` decodes them, which matters for snapshots that
reached disk as a `byte[]` rather than through the SkiaSharp serializer. Removing them
would have been a regression.

So instead of changing the sets, each helper now documents the split — what its serializer
writes, what its comparer can read, and (for `AddImageComparer`) why GIF and ICO are
deliberately absent now that #2180 routes their PNG-stored snapshots to the PNG
registration.

## Not addressed here

The real defect behind that part of the review finding is separate and larger:
`Snapshot.Validate(skImage, SnapshotType.Bmp)` finds no matching serializer, falls through
to `HumanReadableSnapshotSerializer` — which accepts every value
(`HumanReadableSnapshotSerializer.cs:41-45`) — and writes a text dump of the object into a
`.bmp` file, which the registered image comparer then tries to decode. Making an
unsupported (value, type) pair fail loudly instead is a design change; it is filed
separately.

## Verification

```
dotnet build slnx/Meziantou.Framework.SnapshotTesting.slnx /p:TreatsWarningsAsErrors=true          # 0 warnings, 0 errors
dotnet build slnx/Meziantou.Framework.SnapshotTesting.SkiaSharp.slnx /p:TreatsWarningsAsErrors=true # 0 warnings, 0 errors
dotnet build slnx/Meziantou.Framework.SnapshotTesting.ImageSharp.slnx /p:TreatsWarningsAsErrors=true # 0 warnings, 0 errors
dotnet test tests/Meziantou.Framework.SnapshotTesting.Tests --filter-not-class "*EndToEnd*"         # 116/116 passed
```


src/Meziantou.Framework.SnapshotTesting/SnapshotComparerCollectionImageComparerExtensions.cs
```diff
@@ -4,6 +4,18 @@ public static class SnapshotComparerCollectionImageComparerExtensions
...
+        /// </summary>
+        /// <remarks>
+        /// Alternate spellings need no registration of their own: <see cref="SnapshotType.Create(string)" />
+        /// resolves <c>jpg</c> to <see cref="SnapshotType.Jpeg" /> and <c>tif</c> to
+        /// <see cref="SnapshotType.Tiff" />, and snapshot types compare by their canonical name.
+        /// <para>
+        /// GIF and ICO are absent on purpose. Their serializers store PNG frames, and comparers are resolved
+        /// by the format a snapshot is stored in, so those snapshots already use the PNG registration.
+        /// </para>
+        /// </remarks>
```
meziantou: Remove the remarks

src/Meziantou.Framework.SnapshotTesting.SkiaSharp/SnapshotSettingsSkiaSharpExtensions.cs
```diff
@@ -14,6 +14,11 @@ public static class SnapshotSettingsSkiaSharpExtensions
         /// serialized as PNG, JPEG, and WebP snapshots.
         /// </summary>
         /// <param name="settings">Optional image comparison settings. When <see langword="null"/>, exact pixel comparison is used.</param>
+        /// <remarks>
+        /// The serializer writes PNG, JPEG and WebP. The comparer is registered for more formats than that on
+        /// purpose: it decodes through <c>SKCodec</c>, so it can also compare BMP, GIF and ICO snapshots that
+        /// reached disk as a <see cref="byte" /> array rather than through this serializer.
+        /// </remarks>
```
meziantou: Remove the remarks

src/Meziantou.Framework.SnapshotTesting.ImageSharp/SnapsthotSettingsImageSharpExtensions.cs
```diff
@@ -12,6 +12,10 @@ public static class SnapsthotSettingsImageSharpExtensions
         /// Supports <see cref="SixLabors.ImageSharp.Image"/> values serialized as PNG, JPEG, BMP, TIFF, and WebP snapshots.
         /// </summary>
         /// <param name="settings">Optional image comparison settings. When <see langword="null"/>, exact pixel comparison is used.</param>
+        /// <remarks>
+        /// The serializer writes PNG, JPEG, BMP, TIFF and WebP, and the comparer is registered for the same
+        /// set, which is also what ImageSharp can decode.
+        /// </remarks>
```
meziantou: Remove the remarks

src/Meziantou.Framework.SnapshotTesting/SnapshotComparerCollectionImageComparerExtensions.cs
meziantou: Removed in 70776fe.

src/Meziantou.Framework.SnapshotTesting.SkiaSharp/SnapshotSettingsSkiaSharpExtensions.cs
meziantou: Removed in 70776fe.

src/Meziantou.Framework.SnapshotTesting.ImageSharp/SnapsthotSettingsImageSharpExtensions.cs
meziantou: Removed in 70776fe.

PR #2192, opened 2026-08-27
Give each thread its own NullabilityInfoContext

meziantou:
`PublicApi.Generate(Assembly, …)` shared one `NullabilityInfoContext` across every caller in the process, so concurrent calls could throw.

## The defect

`PublicApiModelBuilder` held a single static instance:

```csharp
private static readonly NullabilityInfoContext NullabilityInfoContext = new();
```

Every nullability lookup on the reflection path went through it — 11 call sites covering fields, properties, events, parameters, return types and extension receivers. The .NET documentation for the type is explicit:

> This type is not thread safe. The `Create` methods can't be called concurrently from multiple threads, as they use a non-concurrent cache internally. Concurrent calls can result in an `InvalidOperationException`.

`PublicApi` is a public static API with no documented thread affinity, so concurrent use is a reasonable expectation — a test suite verifying several assemblies at once, or a host generating for several assemblies in parallel. Calling it that way produced:

```
System.InvalidOperationException : Operations that change non-concurrent collections must have exclusive access.
A concurrent update was performed on this collection and corrupted its state.
System.InvalidOperationException : An item with the same key has already been added. Key: Demo.Sample
```

The metadata path (`Generate(string assemblyPath, …)`) does not use this type, which is why the MSBuild task never hit it.

## The change

No caching at all: each call site creates its own `NullabilityInfoContext`. There is no shared nullability state left in the class.

Caching only existed to avoid re-allocating the context, and that turns out not to be worth anything. Five sequential `Generate(Assembly)` calls over `Meziantou.Framework.dll` (437 KB), warm JIT:

| | per-call (ms) |
|---|---|
| shared instance | 17, 16, 15, 14, 14 |
| new instance per call | 13, 12, 12, 12, 12 |

Slightly faster without, presumably because a fresh context's dictionary stays empty rather than accumulating entries across the whole assembly.

## Verification

- New test `Generate_FromReflection_SupportsConcurrentCalls` runs 64 concurrent `Generate(Assembly)` calls over a type with nullable properties, arrays, generics, events and `out` parameters, then asserts every result matches a sequential run.
- The concurrent calls are deliberately the *first* ones: an earlier version of the test computed the expected output first, which warmed the shared context's cache and hid the race entirely (it passed on `main` 3/3). With a cold cache it fails on a shared instance 3/3 with the exception above, and passes 3/3 here.
- That test is also what prevents the shared instance from coming back.
- Full suite: 101/101 passing.
- `dotnet run ./eng/update-all.cs` produced no changes.

## Note

The process-wide `EnumMetadataCache` in `PublicApiModelReader` raises a related question — it is lock-guarded so it is not a correctness problem here, but it is keyed on bare type names with no assembly identity and caches negative results for the life of the process. That is a separate finding and is not addressed here.

Found during a review of the PublicApiGenerator projects.


src/Meziantou.Framework.PublicApiGenerator/PublicApiModelBuilder.cs
```diff
@@ -6,7 +6,13 @@ namespace Meziantou.Framework.PublicApiGenerator;
 
 internal static class PublicApiModelBuilder
 {
-    private static readonly NullabilityInfoContext NullabilityInfoContext = new();
+    // NullabilityInfoContext is not thread safe: its Create methods use a non-concurrent cache internally and
```
meziantou: Not the right fix. Remove caching completely and instantiate a new NullabilityInfoContext whenever it's needed.

meziantou: Done in d9a4d36f5 — the cache is gone entirely, and each of the 11 call sites now does `new NullabilityInfoContext().Create(...)`. No shared state of any kind is left in `PublicApiModelBuilder` for nullability.

I also checked the cost, since that was the only reason to cache: it isn't one. Five sequential `Generate(Assembly)` calls over `Meziantou.Framework.dll` (437 KB), warm JIT — **12, 12, 12, 12 ms** without the cache versus **17, 16, 15, 14 ms** with the shared instance. Slightly faster without, presumably because a fresh context's dictionary stays empty instead of accumulating entries.

`Generate_FromReflection_SupportsConcurrentCalls` is what keeps this from regressing: it fails on a shared instance (3/3 runs, `Operations that change non-concurrent collections must have exclusive access`) and passes here (3/3). Full suite 101/101.

PR #2194, opened 2026-08-27
Emit the last container log line when it has no trailing newline

meziantou:
`ReadMultiplexedLogsAsync` accumulates frame text per stream and only emits on `\n` (`TryReadLine`). When the stream ended, whatever remained in the buffer was discarded rather than flushed.

A container whose readiness marker is written without a trailing newline — a `printf` without `\n`, or a process that dies mid-line — therefore never produced that entry. `Wait.ForLogMessage` then waits out the full `StartupTimeout` and reports a timeout for a message that was actually printed.

## What changed

After the frame loop completes, any non-empty remainder of each stream's buffer is emitted as a final `LogEntry`, through the same `ParseLogLine` as every other line, so the timestamp prefix is still stripped and preserved.

`ReadMultiplexedLogsAsync` becomes `internal` so it can be tested directly. It is pure stream-in / entries-out, which makes it worth pinning: this decoder had no coverage at all.

No public API change, so `ref/` is unchanged.

## Verification

New `DockerApiRuntimeTests` builds synthetic Docker frames (8-byte header, big-endian payload length) and covers four cases:

- `SplitsFramesIntoLines` — multiple lines, stdout and stderr routed correctly.
- `JoinsALineSplitAcrossFrames` — one line arriving in two frames.
- `YieldsTheTrailingLineWithoutANewline` — the fix.
- `KeepsTheTimestampOfTheTrailingLine` — the trailing line goes through the same parsing as the rest.

With the flush removed and everything else identical, the two trailing-line tests fail and the other two pass, so they pin this behavior specifically. 4/4 green with the fix.

New test file rather than an addition to an existing one: there was no test class for `DockerApiRuntime`, and folding stream-decoding tests into `DockerApiCreateRequestBuilderTests` would have misnamed both. Happy to merge them if you'd rather.

### DIFF
--- a/src/Meziantou.Framework.TemporaryContainers/ContainerRuntimes/DockerApiRuntime.cs
+++ b/src/Meziantou.Framework.TemporaryContainers/ContainerRuntimes/DockerApiRuntime.cs
@@ -328,6 +328,8 @@ internal sealed class DockerApiRuntime : ContainerRuntime
             if (buffer.Length == 0)
                 continue;
 
+            // TrimEnd mirrors what TryReadLine does for every other line, so the last one is not the only one that can
+            // come out with a trailing CR. It matters when the stream stops between the CR and the LF of a CRLF ending.
             var parsed = ParseLogLine(logStream, buffer.ToString().TrimEnd('\r'));
             if (parsed is not null)
                 yield return parsed;
--- a/src/Meziantou.Framework.TemporaryContainers/ContainerRuntimes/DockerApiRuntime.cs
+++ b/src/Meziantou.Framework.TemporaryContainers/ContainerRuntimes/DockerApiRuntime.cs
@@ -299,7 +299,7 @@ internal sealed class DockerApiRuntime : ContainerRuntime
         return new InvalidOperationException(message);
     }
 
-    private static async IAsyncEnumerable<LogEntry> ReadMultiplexedLogsAsync(Stream stream, [EnumeratorCancellation] CancellationToken cancellationToken)
+    internal static async IAsyncEnumerable<LogEntry> ReadMultiplexedLogsAsync(Stream stream, [EnumeratorCancellation] CancellationToken cancellationToken)
     {
         var buffers = new Dictionary<LogStream, StringBuilder>
         {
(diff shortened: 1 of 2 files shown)


src/Meziantou.Framework.TemporaryContainers/ContainerRuntimes/DockerApiRuntime.cs
```diff
@@ -319,6 +319,19 @@ private static async IAsyncEnumerable<LogEntry> ReadMultiplexedLogsAsync(Stream
...
+
+        // The stream ended. Whatever is left was never terminated by a newline, but it is still a log line: a process
+        // that writes its readiness marker with 'printf' and no '\n', or that dies mid-line, would otherwise never be
+        // reported and every wait strategy watching for it would time out.
+        foreach (var (logStream, buffer) in buffers)
+        {
+            if (buffer.Length == 0)
+                continue;
+
+            var parsed = ParseLogLine(logStream, buffer.ToString().TrimEnd('\r'));
```
meziantou, line 333: Why does it need to trim the `\r`?

meziantou: Strictly speaking it doesn't *need* to — it's a no-op on almost every input. I measured which ones it actually changes by running the same frames with and without it:

| final bytes on the stream | with `TrimEnd` | without |
|---|---|---|
| `SERVER READY` | `SERVER READY` | `SERVER READY` |
| `SERVER READY\r\n` | `SERVER READY` | `SERVER READY` |
| `line one\r\nSERVER READY` | `line one`, `SERVER READY` | same |
| **`SERVER READY\r`** | `SERVER READY` | `SERVER READY␍` |
| **`50%\r100%\r`** | `50%␍100%` | `50%␍100%␍` |

So it only does anything when the stream stops on a bare CR: a container writing CRLF endings that gets killed between the CR and the LF, or a final progress line that redraws with `\r` and never terminates.

The reason I kept it is consistency rather than necessity. `TryReadLine` already does `TrimEnd('\r')` on every complete line, so without it here the *last* line becomes the only one that can come back with a trailing CR — which would bite anyone doing `entry.Message == "SERVER READY"` rather than a substring match, and only in the truncated case, which is a nasty thing to debug.

I've pushed two things so this isn't just an unexplained call:

- a comment stating the case it covers;
- `ReadMultiplexedLogsAsync_DoesNotEndTheTrailingLineWithACarriageReturn`, which pins it. I checked it actually fails with the `TrimEnd` removed and everything else identical, so it's pinning this line specifically and not passing by accident.

That said, it is a narrow case and I'm not attached to it — if you'd rather not carry speculative handling, say the word and I'll drop the trim and the test in a follow-up commit. Leaving the thread open for you.

PR #2202, opened 2026-08-27
Correct the documented PostgreSQL image tag

meziantou:
The XML doc on `CreatePostgreSql()` says the default image is `postgres:17`; the line below it uses `postgres:18`. The doc is what shows up in IntelliSense and on the API reference, so it's the version most people will read.

I checked the other three helpers while I was here — Redis (`redis:8.2`), MongoDB (`mongo:8`) and SQL Server (`mcr.microsoft.com/mssql/server:2025-latest`) all match their code. Only PostgreSQL had drifted.

## Verification

Doc comment only; `ref/` carries signatures without XML docs, so it is unchanged. Builds clean on both target frameworks with `/p:TreatWarningsAsErrors=true`.

Note: the `postgres:16` in `readme.md` is **not** drift — it illustrates the `CreatePostgreSql(ImageSource)` overload for overriding the image, not the default. Left alone.

### DIFF
--- a/src/Meziantou.Framework.TemporaryContainers/Containers/Postgresql/ContainerDefinitionPostgreSqlExtensions.cs
+++ b/src/Meziantou.Framework.TemporaryContainers/Containers/Postgresql/ContainerDefinitionPostgreSqlExtensions.cs
@@ -8,7 +8,7 @@ public static class ContainerDefinitionPostgreSqlExtensions
     extension(ContainerDefinition)
     {
         /// <summary>Creates a definition pre-configured for a PostgreSQL container (port 5432, a default password, and a readiness wait strategy).</summary>
-        /// <returns>A PostgreSQL container definition using the <c>postgres:17</c> image.</returns>
+        /// <returns>A PostgreSQL container definition using the <c>postgres:18</c> image.</returns>
         public static PostgreSqlContainerDefinition CreatePostgreSql()
         {
             return CreatePostgreSql(ImageSource.FromRegistry("postgres:18"));
--- a/renovate.json5
+++ b/renovate.json5
@@ -3,6 +3,23 @@
   "extends": [
     "github>meziantou/renovate-config"
   ],
+  "customManagers": [
+    {
+      // The shared preset only matches the ImageSource.FromRegistry(...) call, so the image named in the XML
+      // documentation right above it was left behind and drifted (postgres was documented as 17 while the code
+      // created an 18 container). Anchored on the 'using the <c>image:tag</c> image' wording so it cannot match
+      // the illustrative <c>redis:8</c> samples or <c>host:port</c> elsewhere in the documentation.
+      "customType": "regex",
+      "description": "Default container images named in the XML documentation of the ContainerDefinition factory methods",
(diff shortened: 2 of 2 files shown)


src/Meziantou.Framework.TemporaryContainers/Containers/Postgresql/ContainerDefinitionPostgreSqlExtensions.cs
```diff
@@ -8,7 +8,7 @@ public static class ContainerDefinitionPostgreSqlExtensions
     extension(ContainerDefinition)
     {
         /// <summary>Creates a definition pre-configured for a PostgreSQL container (port 5432, a default password, and a readiness wait strategy).</summary>
-        /// <returns>A PostgreSQL container definition using the <c>postgres:17</c> image.</returns>
+        /// <returns>A PostgreSQL container definition using the <c>postgres:18</c> image.</returns>
```
meziantou, line 11: Can you make sure Renovate also updates this version

meziantou: Done — added a `customManagers` entry to `renovate.json5`.

The reason it drifted: the `.cs` manager in `github>meziantou/renovate-config` matches only the call site —

```
ImageSource\.FromRegistry\("(?<depName>.+):(?<currentValue>[^"@]+)"\)
```

so `3aea2be4a` (`chore(deps): update all-dependencies` #1830) bumped `FromRegistry("postgres:17")` → `18` and left the `<c>postgres:17</c>` on the line above untouched.

The new manager matches the documentation wording instead:

```
using the <c>(?<depName>[^<]+):(?<currentValue>[^<@]+)</c> image
```

I anchored it on `using the … image` rather than a bare `<c>x:y</c>` because a loose pattern would also match `<c>host:port</c>` in `RedisContainer` and the illustrative `<c>redis:8</c>` samples in `ContainerDefinition` and `ImageSource`, none of which are default images.

Verified before pushing:

- `renovate-config-validator --strict` passes. I sanity-checked the validator by feeding it a bogus key first, and it did reject it — so the run is meaningful, not vacuous.
- `managerFilePatterns` selects exactly the four `Containers/*Extensions.cs` files, and the regex finds all four with the right values, including `mcr.microsoft.com/mssql/server` : `2025-latest` (greedy `[^<]+` binds to the last colon).
- No false positives across every `.cs` file in the repo.
- Each documented image extracts the **same** `depName:currentValue` as the preset's code manager already extracts from the same file — `mongo:8`, `postgres:18`, `redis:8.2`, `mcr.microsoft.com/mssql/server:2025-latest` — so both occurrences are the same dependency and land in one PR rather than two.

One caveat worth knowing: this is regex-on-prose, so rewording that sentence would silently stop the matching and the drift could come back unnoticed. If you want a hard guard, a test asserting the generated XML doc agrees with `ContainerDefinition.CreateX().Image` would catch it regardless of Renovate. Happy to add that here or in a follow-up — I left it out to keep this PR to the one-line doc fix plus the config.

PR #2278, opened 2026-08-28
Implement the UTS 39 confusable skeleton algorithm

meziantou:
Implements the UTS #39 confusable detection algorithm, per @meziantou's review comment on
#2108 ("Implement the right algorithm"). Replaces #2108, which is closed.

The package cited UTS #39 but did not implement it. The spec defines

> skeleton(X) = toNFD(toConfusable(toNFD(X)))

and two strings are confusable when their skeletons are equal. `ReplaceConfusablesCharacters`
applies only the middle step, to un-normalized input, so callers doing
`Replace(a) == Replace(b)` — the only comparison the package made possible — got false
negatives on any input where normalization matters. Verified against the shipped library:

| case | `Replace(a) == Replace(b)` | `AreConfusable(a, b)` |
|---|---|---|
| `café` composed vs decomposed | ❌ false | ✅ true |
| `café` vs Cyrillic `е` + combining acute | ❌ false | ✅ true |
| `Ï` vs `I` + U+0308 | ❌ false | ✅ true |
| `Zoë` vs Cyrillic `ё` | ❌ false | ✅ true |
| `paypal` vs Cyrillic `раураl` | ✅ true | ✅ true |
| `paypal` vs `example` | ✅ false | ✅ false |

Every one of those false negatives is a homograph that slips past a naive check while looking
identical to a human.

## What changed

Two additive public methods; no existing behaviour is modified.

- `Unicode.GetConfusableSkeleton(string)` — NFD, then the confusable mapping, then NFD again.
- `Unicode.AreConfusable(string, string)` — skeleton equality, with an ordinal fast path for
  identical strings.

**Why a single mapping pass is correct.** Older descriptions of the algorithm apply the
mapping repeatedly until it stabilizes. That is unnecessary here, and I checked rather than
assumed: of the 6,565 entries in the shipped table, **0** have a replacement containing a
character that is itself a mapping source. The table is closed, so one pass reaches the fixed
point.

**Why the table must keep its ASCII entries.** This is what closed #2108. 384 characters in
U+0020..U+2FFF are not mapped themselves but decompose to something that is — `Ï` (U+00CF)
has NFD `I` + U+0308, and `I` → `l` is one of the eight ASCII mappings #2108 deleted. So
`GetConfusableSkeleton("Ï")` is `l` + U+0308, and removing those entries would silently break
it. There is a test pinning exactly this.

**Ill-formed input.** `string.Normalize` throws `ArgumentException` on unpaired surrogates, so
these methods do too. That is documented on both, and preferable to silently reducing input
that a caller believed was validated.

## Not in this PR

`ReplaceConfusablesCharacters` is untouched, so the ASCII text-mangling issue
(`"Item 1 of 10"` → `"ltern l of lO"`) is still open. That is a separate decision about what
that method is for, deliberately not settled by this PR.

Mixed-script detection — the other half of UTS #39 — needs Unicode Script property data the
package does not currently ship, and follows separately.

## Verification

`dotnet test ... /p:TreatsWarningsAsErrors=true` — **88 passed**, 0 failed.

New facts cover the four confusable pairs above, three non-confusable pairs, the
`Ï` → `l` + U+0308 fold that pins the ASCII dependency, skeleton idempotence, the empty and
identical-string paths, and argument validation including the unpaired-surrogate throw.

`ref/Meziantou.Framework.Unicode.cs` is regenerated and included.


src/Meziantou.Framework.Unicode/Unicode.cs
```diff
@@ -150,4 +150,114 @@ public static bool TryGetCharacterInfo(char value, out UnicodeCharacterInfo info
...
+    private static bool GetIsGlobalizationInvariant()
+    {
+        // Mirrors how the runtime resolves the setting: the switch set by the InvariantGlobalization
+        // MSBuild property takes precedence over the environment variable.
+        if (AppContext.TryGetSwitch("System.Globalization.Invariant", out var isEnabled))
+            return isEnabled;
+
+        var value = Environment.GetEnvironmentVariable("DOTNET_SYSTEM_GLOBALIZATION_INVARIANT");
+        return value is "1" || string.Equals(value, "true", StringComparison.OrdinalIgnoreCase);
+    }
```
meziantou: There is already the logic somewhere (maybe in the xunit project). Can you extract it to the common folder and import it in both projects

meziantou: Good catch — the logic was already in `TestEnvironment.IsGlobalizationInvariant()` in the Xunit project, and I duplicated it instead of reusing it.

Extracted to `common/GlobalizationHelper.cs` and linked into both projects with `<Compile Include="../../common/GlobalizationHelper.cs" />`, matching how `SharedHttpClient.cs` and `TypeNameHelper.cs` are shared. `TestEnvironment.IsGlobalizationInvariant()` now delegates to it, so its public signature is unchanged and no `ref/` file moves.

Verified the shared implementation still drives the `RunIf` machinery correctly, in this PR and in an unrelated consumer:

- Unicode tests — ICU: 176 passed / 2 skipped; invariant: 158 passed / 20 skipped
- Scheduling tests (uses `[RunIf(globalizationMode: TestGlobalizationMode.NotInvariant)]`) — ICU: 524 passed / 0 skipped; invariant: 450 passed / 12 skipped

The skip counts flipping correctly in both directions is what proves the delegation works, rather than the helper just always returning one value.

One thing I left alone: there is a third copy in `src/Meziantou.Framework/GlobalizationMode.cs`, which reads the same switch via `AppContextConfigHelper` and also handles `PredefinedCulturesOnly`. Folding that in would touch an unrelated package and a second concern, so I kept it out of this PR — happy to do it separately if you want all three unified.

PR #2304, opened 2026-08-28
Escape the address and number segments of URI payloads

meziantou:
## The problem

`Email`, `Phone`, `Sms` (`QRCodePayload.cs:132-190`) and `Bitcoin` (`:399-432`) escape their
**optional parameters** with `Uri.EscapeDataString` but append the **primary value** raw:

```
Email("victim@x.com?bcc=attacker@evil.example", "Hi")
  → mailto:victim@x.com?bcc=attacker@evil.example?subject=Hi

Bitcoin("addr?amount=999", amount: 0.05m)
  → bitcoin:addr?amount=999?amount=0.05

Sms("+1555", "a:b")
  → smsto:+1555:a:b
```

Mail clients parse the first `?`, so the injected `bcc` takes effect and the intended
`subject=Hi` is swallowed into its value — a QR code that silently blind-copies a third party
on the user's message.

`Sms` fails on ordinary input too: the format is `smsto:<number>:<message>`, so a `:` in the
number shifts the split.

## The change

Percent-encode the primary segment of each payload, preserving the characters each scheme
needs:

- **mail addresses** keep `@` readable (`%40` restored), so `test@example.com` is unchanged
- **phone numbers** keep the leading `+` that `tel:` requires
- **bitcoin addresses** are base58/bech32, so encoding is a no-op for any valid address

Valid inputs produce byte-identical output to before; only structurally significant
characters change. The SMS *message* is left alone — it is the trailing free text, so a `:`
in it is harmless.

## Verification

319 tests pass on net10.0 and net11.0. All thirteen existing assertions for these four
builders pass unchanged, which is the point: this only affects inputs that were already
producing a malformed URI.

New coverage: query-parameter injection through a mail address and a bitcoin address, `#`/`&`/
whitespace in a mail address, and a `:` in an SMS number.

### DIFF
--- a/src/Meziantou.Framework.QRCode/QRCodePayload.cs
+++ b/src/Meziantou.Framework.QRCode/QRCodePayload.cs
@@ -202,7 +202,15 @@ public static class QRCodePayload
     /// </remarks>
     private static string EscapeMailAddress(string address)
     {
-        return Uri.EscapeDataString(address).Replace("%40", "@", StringComparison.Ordinal);
+        // An addr-spec has a single '@' between the local part and the domain. Only that one is
+        // left readable; any other '@' is not a separator, so it stays encoded.
+        var separatorIndex = address.AsSpan().LastIndexOf('@');
+        if (separatorIndex < 0)
+        {
+            return Uri.EscapeDataString(address);
+        }
+
+        return Uri.EscapeDataString(address[..separatorIndex]) + "@" + Uri.EscapeDataString(address[(separatorIndex + 1)..]);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Meziantou.Framework.QRCode/QRCodePayload.cs
```diff
@@ -181,12 +181,36 @@ public static string Sms(string number, string? message = null)
...
+    {
+        return Uri.EscapeDataString(address).Replace("%40", "@", StringComparison.Ordinal);
+    }
+
+    /// <summary>
+    /// Percent-encodes a phone number, keeping the leading <c>+</c> that <c>tel:</c> needs.
+    /// </summary>
+    private static string EscapePhoneNumber(string number)
+    {
+        return Uri.EscapeDataString(number).Replace("%2B", "+", StringComparison.Ordinal);
```
meziantou: I'm not sure the Replace is ok if we only intent to replace the first character

meziantou: Good catch - the Replace restored every occurrence while the doc comment claimed it kept only the leading one. Fixed in the latest commit by replacing both Replace calls with positional handling:

- `EscapePhoneNumber` keeps a `+` only when it is the first character (RFC 3966 allows it only as the international prefix), so `+1+555` now gives `tel:+1%2B555`.
- `EscapeMailAddress` keeps only the last `@`, the one separating the local part from the domain, so `a@b@c.com` gives `mailto:a%40b@c.com`.

This also tightened the injection case: `Email("victim@x.com?bcc=attacker@evil.example")` used to leave both `@` readable and now yields `mailto:victim%40x.com%3Fbcc%3Dattacker@evil.example` with a single separator. Added theories pinning both rules.

PR #2323, opened 2026-08-29
Keep the torrent keys the model does not represent when writing

meziantou:
**Stack 2 of 2 · merges into `fix/bencode-info-hash-source-bytes` (#2322) · must merge after it.**

## The problem

`TorrentFile` models six top-level keys and `TorrentInfo` six info keys, and both rebuilt their dictionary purely from those properties (`TorrentFile.cs:162-216`, `TorrentInfo.cs:94-134`). Everything else the file carried was read past at parse time and absent at write time: `url-list` (web seeds, BEP 19), `httpseeds`, `nodes` (DHT bootstrap, BEP 5), `encoding`, and inside `info` — `source`, `md5sum`, `name.utf-8`, `attr`, the BitTorrent v2 keys.

The readme presents exactly this as the supported workflow — parse `file.torrent`, `WriteToAsync` to `copy.torrent`. The "copy" is not a copy: it has lost its web seeds and DHT nodes, and before #2322 it also had a different info-hash.

## The change

Parsing keeps the dictionary each object came from. On write, `TorrentDictionaryMerge.Merge` walks the source in order and, for each entry:

- a key the model owns and currently sets → the model's value wins,
- a key the model owns but no longer sets → dropped, so clearing a property removes the key,
- anything else → carried over verbatim.

Modelled keys the source did not contain are appended afterwards. A torrent constructed in memory has no source, so it writes exactly what it did before.

One asymmetry worth knowing: an explicit `private 0` parses to `IsPrivate = false` and is therefore written as *absent* rather than as `0`. Those two are semantically identical, and since #2322 the info-hash comes from the parsed bytes, so this no longer affects the torrent's identity.

## Verification

`dotnet build /p:TreatWarningsAsErrors=true` clean; `dotnet test` green — 130 tests across net10.0 and net11.0.

`ToUtf8ByteArray_PreservesKeysTheModelDoesNotRepresent` fails against the layer below and passes here:

```
failed ToUtf8ByteArray_PreservesKeysTheModelDoesNotRepresent (7ms)
```

It parses a torrent carrying `url-list`, `nodes`, `source` and `md5sum`, writes it, and asserts all four survive and that the round-tripped info-hash matches the original. Four more tests cover the rest of the contract: modelled fields still override the parsed value, a cleared field is removed, a constructed torrent writes only modelled keys, and the ubuntu fixture keeps `e1fc140a6391357fa1cf08ddb70274f9c05eb88b` through a full write/re-parse cycle.

Found by a project review of `src/Meziantou.Framework.Bencode`.

### DIFF
--- a/src/Meziantou.Framework.Bencode/BencodeDecoder.cs
+++ b/src/Meziantou.Framework.Bencode/BencodeDecoder.cs
@@ -110,6 +110,7 @@ internal static class BencodeDecoder
     private static BencodeDictionary ParseDictionary(ReadOnlySpan<byte> data, ref int index, int depth)
     {
         EnsureDepth(depth);
+        var start = index;
         index++; // d
         var result = new BencodeDictionary();
 
@@ -121,6 +122,8 @@ internal static class BencodeDecoder
             if (data[index] == (byte)'e')
             {
                 index++;
+                result.SourceOffset = start;
+                result.SourceLength = index - start;
                 return result;
             }
 
--- a/src/Meziantou.Framework.Bencode/BencodeDictionary.cs
+++ b/src/Meziantou.Framework.Bencode/BencodeDictionary.cs
@@ -23,6 +23,12 @@ public sealed class BencodeDictionary : BencodeValue, IReadOnlyDictionary<Bencod
 
     public override BencodeValueKind Kind => BencodeValueKind.Dictionary;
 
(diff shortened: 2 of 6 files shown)


meziantou:
Closing in favour of modelling these fields explicitly on `TorrentFile`/`TorrentInfo` rather than carrying unknown keys through the write path — see the replacement PR.

The tradeoff being accepted: keys outside the modelled set are still dropped when a parsed torrent is written back out. #2322 already landed, so the info-hash comes from the parsed bytes and is unaffected either way.

PR #2325, opened 2026-08-29
Keep the AMSI context alive until its sessions are closed

meziantou:
**Stack 1 of 4 · merges into `main` · must merge before the rest of the stack.**

## The finding

`AmsiSessionSafeHandle.ReleaseHandle` hands the context back to `AmsiCloseSession`:

```csharp
// Internals/AmsiSessionSafeHandle.cs:26-31
protected override bool ReleaseHandle()
{
    Debug.Assert(Context is not null);
    Amsi.AmsiCloseSession(Context, handle);   // -> Context.DangerousGetHandle()
    return true;
}
```

Nothing kept that context alive. The `IsInvalid` override on line 24 was meant to guard it, but it cannot: `SafeHandle.IsInvalid` reads the raw `handle` field, and `SafeHandle` never clears that field on dispose — it only flips `IsClosed`.

I reproduced it with a faithful copy of both handle classes on .NET 10:

```
-- user writes: context.Dispose(); session.Dispose(); --
  CTX released -> AmsiUninitialize(0x1000)
  after ctx.Dispose(): ctx.IsClosed=True, ctx.IsInvalid=False, sess.IsInvalid=False
  SESSION released -> AmsiCloseSession(ctx=0x1000, ...) with ctx.IsClosed=True   <-- freed
```

Two triggers, both plain user code:

```csharp
var context = AmsiContext.Create("app");
var session = context.CreateSession();
context.Dispose();
session.Dispose();          // AmsiCloseSession on an uninitialized context

context.Dispose();
session.IsMalware(payload, "x");   // AmsiScanString on an uninitialized context
```

An uninitialized `HAMSICONTEXT` handed back to `amsi.dll` is undefined behaviour inside the provider. The existing tests use `using` in reverse order, which happens to be the safe order, so nothing caught this.

Finalization was **not** affected — `AmsiSessionSafeHandle.Context` keeps the context reachable until the session finalizer has run, which I verified rather than assumed. Only explicit dispose ordering was broken.

## What changed

- `AmsiSessionSafeHandle` takes a reference on the context with `DangerousAddRef` and releases it in `ReleaseHandle`. `context.Dispose()` now defers the real `AmsiUninitialize` until the last session is closed.
- The `IsInvalid` override is gone — it was buying nothing.
- The scan helpers in `Internals/Amsi.cs` bracket their native calls with `DangerousAddRef`/`DangerousRelease`. They read raw pointers out of the handles, so the handles are unreferenced in the frame for the duration of the call and a collection could otherwise run the critical finalizer mid-scan.
- Dropped the dead `session.Context = _handle;` in `CreateSession` — the constructor already set it, and the property is now a private field, so it had to go with this change.

## Verification

- `dotnet run ./eng/update-all.cs` produces no further changes.
- Verified empirically that `DangerousAddRef` succeeds on the zero-handle `DefaultSession` (it is on the hot path of every `AmsiContext.IsMalware` call) and that ref-counting defers `ReleaseHandle` past `Dispose`.
- 3 regression tests added, using `[RunIf(TestOperatingSystems.Windows)]` so they genuinely run on the Windows CI leg rather than being skipped like the existing EICAR tests.

⚠️ **The new tests have never been executed.** This work was done on macOS, where all AMSI tests fail with `DllNotFoundException` on `amsi.dll` (pre-existing — the existing tests only `SkipIf` GitHub Actions, not non-Windows). The Windows CI leg is their first real run.

### DIFF
--- a/src/Meziantou.Framework.Win32.Amsi/AmsiContext.cs
+++ b/src/Meziantou.Framework.Win32.Amsi/AmsiContext.cs
@@ -44,7 +44,6 @@ public sealed class AmsiContext : IDisposable
     public AmsiSession CreateSession()
     {
         var result = Amsi.AmsiOpenSession(_handle, out var session);
-        session.Context = _handle;
         if (result != 0)
             throw new Win32Exception(result);
 
--- a/src/Meziantou.Framework.Win32.Amsi/Internals/Amsi.cs
+++ b/src/Meziantou.Framework.Win32.Amsi/Internals/Amsi.cs
@@ -27,28 +27,74 @@ internal static partial class Amsi
 
     internal static int AmsiOpenSession(AmsiContextSafeHandle amsiContext, out AmsiSessionSafeHandle session)
     {
-        var result = PInvoke.AmsiOpenSession((HAMSICONTEXT)amsiContext.DangerousGetHandle(), out var nativeSession);
-        session = new AmsiSessionSafeHandle(amsiContext, (nint)nativeSession);
-        return result;
+        var contextAddRef = false;
+        try
+        {
+            amsiContext.DangerousAddRef(ref contextAddRef);
+            var result = PInvoke.AmsiOpenSession((HAMSICONTEXT)amsiContext.DangerousGetHandle(), out var nativeSession);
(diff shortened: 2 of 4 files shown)


meziantou:

PR #2326, opened 2026-08-29
Treat AMSI return values as HRESULT instead of Win32 error codes

meziantou:
**Stack 2 of 4 · merges into `fix/amsi-handle-lifetime` (#2325) · merge after #2325.**

## The finding

Every AMSI function returns an `HRESULT` — confirmed both in the [docs](https://learn.microsoft.com/en-us/windows/win32/api/amsi/nf-amsi-amsiscanstring) ("If this function succeeds, it returns **S_OK**. Otherwise, it returns an **HRESULT** error code") and in the CsWin32-generated signatures. The wrapper widened that to `int` and did this at all six call sites:

```csharp
if (result != 0)
    throw new Win32Exception(result);
```

Two separate defects:

1. **Wrong success test.** HRESULT success is `result >= 0`, not `result == 0`. A non-`S_OK` success such as `S_FALSE` would throw on a call that actually succeeded, discarding a valid scan result.
2. **Wrong error domain.** `Win32Exception(int)` interprets its argument as a `GetLastError()` code, so `NativeErrorCode` ends up holding an HRESULT and callers cannot branch on it.

CI on #2325 produced a live example of this. A scan on a hosted agent failed and surfaced as:

```
System.ComponentModel.Win32Exception : The device is not ready.
   at AmsiContext.IsMalware(String payload, String contentName) in AmsiContext.cs:line 62
```

AMSI returned `HRESULT_FROM_WIN32(ERROR_NOT_READY)` = `0x80070015`. To be precise about what does and does not go wrong here: `FormatMessage` *did* resolve that value, so the message text is correct — that part of my original description was overstated. What is wrong is `NativeErrorCode`, which holds `0x80070015` (2147942421) rather than `21`, so `catch (Win32Exception e) when (e.NativeErrorCode == ERROR_NOT_READY)` does not match. And the message only survives because this HRESULT happens to be `FACILITY_WIN32`; a provider-specific or `FACILITY_ITF` HRESULT has no system message-table entry and degrades to `Unknown error (0x…)`.

This is the entire error-reporting surface of the package, and AMSI failing is not exotic — no registered provider, AMSI disabled by policy, or a provider rejecting the content all arrive here.

## What changed

- The interop helpers return CsWin32's `HRESULT` instead of widening to `int`, and the public methods call `.ThrowOnFailure()`. This matches the convention already used in the repo at `src/Meziantou.Framework.FullPath/FullPath.cs:731`; every other `Win32Exception` in the repo is paired with `Marshal.GetLastWin32Error()`, so AMSI was the outlier.
- `AmsiInitialize`/`AmsiOpenSession` no longer wrap the `out` parameter in an owning `SafeHandle` when the call failed — on failure it carries no handle that may legally be passed back to `AmsiUninitialize`/`AmsiCloseSession`.

## Behaviour change for callers

Failures now throw what `Marshal.ThrowExceptionForHR` produces — `COMException` for provider HRESULTs, or a mapped BCL type (`E_INVALIDARG` → `ArgumentException`, `E_OUTOFMEMORY` → `OutOfMemoryException`) — instead of `Win32Exception`. The `<exception>` docs were updated to `COMException`.

This is a breaking change for anyone catching `Win32Exception`. The package is at `3.0.0`; flagging it in case you want a major bump or a release note. I did not touch `<Version>` since `eng/update-versions.cs` owns that.

## Verification

- Public API surface unchanged (`ref/` regenerates identically) — only the thrown exception type differs, which the ref file does not capture.
- No new tests: the change is on error paths that need a failing AMSI provider to reach, which I cannot produce on macOS. Called out in the stack summary as a coverage gap rather than papered over.

### DIFF
--- a/src/Meziantou.Framework.Win32.Amsi/AmsiContext.cs
+++ b/src/Meziantou.Framework.Win32.Amsi/AmsiContext.cs
@@ -1,4 +1,4 @@
-using System.ComponentModel;
+using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 
 namespace Meziantou.Framework.Win32;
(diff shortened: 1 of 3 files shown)


meziantou:

PR #2327, opened 2026-08-29
Throw ObjectDisposedException instead of scanning with a disposed context

meziantou:
**Stack 3 of 4 · merges into `fix/amsi-hresult` (#2326) · merge after #2326.**

## The finding

Neither `Dispose` recorded that it ran, and the scan methods go straight to the interop layer, which reads the raw pointer with `DangerousGetHandle()`. That returns the stale pointer after the handle is closed, so:

```csharp
var context = AmsiContext.Create("app");
context.Dispose();
context.IsMalware("anything", "x");   // AmsiScanString with a freed HAMSICONTEXT
```

No exception, no failure code — the freed pointer just goes into the native call. Use-after-dispose is common enough in caller code (a shared context disposed by one path while another still holds it) that a library is expected to turn it into an `ObjectDisposedException`; here it became undefined native behaviour whose crash points at `amsi.dll` rather than at the caller's bug.

`AGENTS.md` also asks for `ObjectDisposedException.ThrowIf` where applicable.

## What changed

Guards on all four scan methods plus `CreateSession`, and on `AmsiSession` a check that the parent context is still alive.

**Note on why this uses an explicit flag rather than `SafeHandle.IsClosed`:** the obvious implementation is `ObjectDisposedException.ThrowIf(_handle.IsClosed, this)`, and it would be wrong. Layer 1 makes the session hold a reference on the context handle, and I verified that `SafeHandle.IsClosed` stays **`false`** after `Dispose()` while any reference is outstanding:

```
after Dispose while ref held: IsClosed=False
  >> ReleaseHandle -> AmsiUninitialize(0x1000)     <-- only after DangerousRelease
```

So `IsClosed` is not a reliable "was `Dispose` called" signal once ref-counting is in play. `AmsiContext` tracks disposal in a `_disposed` field and exposes it internally for `AmsiSession` to check.

## Verification

- 4 tests added (`[RunIf(TestOperatingSystems.Windows)]`): disposed context, disposed session, session whose context was disposed, and double-dispose.
- Public API surface unchanged.

⚠️ Like the rest of the stack, these tests have not been executed — macOS cannot load `amsi.dll`. The Windows CI leg is their first run.

### DIFF
--- a/src/Meziantou.Framework.Win32.Amsi/AmsiContext.cs
+++ b/src/Meziantou.Framework.Win32.Amsi/AmsiContext.cs
@@ -20,6 +20,16 @@ public sealed class AmsiContext : IDisposable
 
     private static readonly AmsiSessionSafeHandle DefaultSession = new();
 
+    private bool _disposed;
+
+    /// <summary>Gets a value indicating whether this context has been disposed.</summary>
+    /// <remarks>
+    /// Tracked separately from the handle: a session holds a reference on the handle, so
+    /// <see cref="System.Runtime.InteropServices.SafeHandle.IsClosed"/> stays <see langword="false"/>
+    /// after <see cref="Dispose"/> until every session is closed.
+    /// </remarks>
+    internal bool IsDisposed => _disposed;
+
     private AmsiContext(AmsiContextSafeHandle context)
     {
         _handle = context;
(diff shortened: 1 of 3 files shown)


meziantou:

PR #2328, opened 2026-08-29
Expose the raw AMSI scan result so admin-blocked content can be detected

meziantou:
**Stack 4 of 4 · merges into `fix/amsi-objectdisposed` (#2327) · merge last.**

## The finding

The public API is a single `bool`. `AmsiResultIsMalware` correctly implements the documented macro (`>= AMSI_RESULT_DETECTED`, 32768), so results in the `AMSI_RESULT_BLOCKED_BY_ADMIN` range (16384–20479) returned `false`. The enum carrying that distinction was `internal` and the value was discarded, so nothing reached the caller.

The [AMSI_RESULT docs](https://learn.microsoft.com/en-us/windows/win32/api/amsi/ne-amsi-amsi_result) are explicit about that range: results within it "are officially blocked by the admin specified policy. In these cases, the script in question will be blocked from executing."

So an application using this package to decide whether to execute or load scanned content would execute content the machine's administrator has explicitly forbidden. That is a policy bypass introduced by the wrapper — a caller using `amsi.dll` directly sees the value and can act on it.

## What changed

Additive only:

- `AmsiResult` becomes public (moved out of `Internals/`, members renamed to .NET conventions: `Clean`, `NotDetected`, `BlockedByAdminStart`, `BlockedByAdminEnd`, `Detected`).
- `Scan(string, string)` and `Scan(byte[], string)` on both `AmsiContext` and `AmsiSession`, returning the raw result. This also exposes the 1–32767 provider-specific risk score, which was previously discarded too.
- `AmsiResultExtensions` with `IsMalware()`, `IsBlockedByAdmin()` and `ShouldBlock()`.
- The existing `IsMalware` overloads keep their current semantics — they still implement the documented macro — and now route through `Scan` and document that they do not cover admin-blocked content, pointing at `ShouldBlock`.

No existing signature changed; `ref/` regenerated (Release + `IsOfficialBuild`) and the diff is purely additive.

```csharp
using var context = AmsiContext.Create("MyApplication");

if (context.Scan(payload, "upload.ps1").ShouldBlock())
{
    // covers both malware and administrator policy
}
```

## Verification

- `dotnet run ./eng/update-all.cs` produces no further changes.
- **20 tests added that actually pass**, on macOS included. The extension methods are pure functions over the enum, so their tests need no antimalware provider and no Windows — they run on every CI leg. They pin the boundaries: 32767/32768 for `IsMalware`, and 16383/16384/20479/20480 for `IsBlockedByAdmin`.
- One `[RunIf(TestOperatingSystems.Windows)]` test covers `Scan` end-to-end; that one has not been executed here.

This is the first AMSI test coverage in the repository that runs outside the Windows leg.

## Note on merge order

Conflicts with #2329 (`fix/amsi-platform-version`) in `ref/Meziantou.Framework.Win32.Amsi.cs` only. That file is generated — whichever lands second should be resolved by regenerating rather than hand-merging:

```
dotnet build src/Meziantou.Framework.Win32.Amsi/Meziantou.Framework.Win32.Amsi.csproj -c Release /p:IsOfficialBuild=true
```


meziantou:

PR #2330, opened 2026-08-29
Document usage and requirements in the AMSI package readme

meziantou:
**Independent — branches off `main`, merges in any order. Documentation only, no code changes.**

## The finding

`src/Meziantou.Framework.Win32.Amsi/readme.md` was a title and a single link. This is the NuGet landing page, and it carried no usage example, did not state that the package is Windows-only, and did not mention that when no antimalware provider is registered every scan reports the content as not detected — worth knowing before building a security control on top of it.

The XML docs do carry good examples; they are just not where someone evaluating the package looks first.

## What changed

Rewrote the readme in the style of the sibling Win32 packages (`Meziantou.Framework.Win32.Jobs`, `Meziantou.Framework.Win32.AccessToken`): a one-paragraph description, a requirements section, usage for string and buffer scans, a session example, and the existing links kept under "Additional resources".

No code touched, so nothing to build or test.

### DIFF
--- a/src/Meziantou.Framework.Win32.Amsi/readme.md
+++ b/src/Meziantou.Framework.Win32.Amsi/readme.md
@@ -1,5 +1,56 @@
 # Meziantou.Framework.Win32.Amsi
 
-## Additional resources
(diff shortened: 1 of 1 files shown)


meziantou:

PR #2336, opened 2026-08-29
Return Empty instead of throwing for FullPath.FromPath with an empty path

meziantou:
`FullPath.FromPath("")` threw `ArgumentException: The value cannot be an empty string. (Parameter 'path')`, because `Path.GetFullPath("")` throws before `FromPath` reaches its own empty check a few lines below.

That makes `FullPath.GetFolderPath` unsafe on exactly the machines where it matters:

```csharp
// Environment.GetFolderPath returns "" when the folder is not defined on the platform —
// routine for MyMusic, MyVideos, CommonDesktopDirectory in containers and on servers.
var music = FullPath.GetFolderPath(Environment.SpecialFolder.MyMusic);  // threw
```

`FullPath.GetFolderPath` is `FromPath(Environment.GetFolderPath(folder))`, so a documented empty result became an unhandled exception. Code that probed special folders and guarded with `if (dir.Length == 0)` had its guard rendered unreachable. It fails in CI and container environments rather than on the developer's box, so it ships green.

## What changed

`FromPath` returns `Empty` for an empty input. This is the same outcome the method already produced one branch later for a path that *resolves* to nothing (`FromPath("/")` → `Empty`) — an empty input is that same case, it just never got there.

`path is { Length: 0 }` rather than `path.Length == 0` deliberately: a `null` argument still falls through to `Path.GetFullPath` and throws `ArgumentNullException` exactly as before, rather than turning into a `NullReferenceException`.

## Behavior change

This is a public behavior change to a shipped package. `FromPath("")` goes from throwing to returning `Empty`. No test in the repo pinned the throw, and `IsEmpty` already existed as the way to detect this state, but it is worth a deliberate look.

It also resolves the second half of a review finding: MFFP0024 suggests `Environment.GetFolderPath` → `FullPath.GetFolderPath`, which was previously not a safe rewrite for this reason.

## Verification

Test added asserting `FromPath("")`, `FromPath(string.Empty)` and `FullPath.Empty` agree. It fails on `main` with the `ArgumentException` above and passes with the fix. Full `Meziantou.Framework.FullPath` suite green — 164 run, 56 skipped (Windows-only), 0 failed. `dotnet run ./eng/update-all.cs` produces no further changes.

### DIFF
--- a/src/Meziantou.Framework.FullPath/FullPath.cs
+++ b/src/Meziantou.Framework.FullPath/FullPath.cs
@@ -441,6 +441,10 @@ public readonly partial struct FullPath : IEquatable<FullPath>, IComparable<Full
         // A path that resolves to nothing already returns Empty below; an empty input is the same case, and
         // Path.GetFullPath would throw before reaching it. Environment.GetFolderPath returns "" for a folder
         // that is not defined on the platform, which makes this a routine input rather than a caller error.
+        //
+        // Empty rather than CurrentDirectory: "" means "no path", not "the current directory". Resolving it against
+        // the current directory would turn an undefined special folder into a real, writable location. "." is the
+        // spelling for the current directory, and Path.GetFullPath("") throws rather than treating the two alike.
         if (path is { Length: 0 })
             return Empty;
 
--- a/src/Meziantou.Framework.FullPath/FullPath.cs
+++ b/src/Meziantou.Framework.FullPath/FullPath.cs
@@ -435,9 +435,15 @@ public readonly partial struct FullPath : IEquatable<FullPath>, IComparable<Full
     public static FullPath CurrentDirectory() => FromPath(Environment.CurrentDirectory);
 
     /// <summary>Creates a <see cref="FullPath"/> from a string path by converting it to an absolute path.</summary>
-    /// <param name="path">The path to convert. Can be relative or absolute.</param>
+    /// <param name="path">The path to convert. Can be relative or absolute. An empty string returns <see cref="Empty"/>.</param>
     public static FullPath FromPath(string path)
     {
+        // A path that resolves to nothing already returns Empty below; an empty input is the same case, and
+        // Path.GetFullPath would throw before reaching it. Environment.GetFolderPath returns "" for a folder
+        // that is not defined on the platform, which makes this a routine input rather than a caller error.
+        if (path is { Length: 0 })
+            return Empty;
+
         // '\' is a regular file name character on Unix, so a path such as @"\\?\a" is a relative file name there, not a device path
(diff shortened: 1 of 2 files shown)


src/Meziantou.Framework.FullPath/FullPath.cs
```diff
@@ -435,9 +435,15 @@ public static unsafe FullPath GetKnownFolderPath(KnownFolder knownFolder)
...
     /// <summary>Creates a <see cref="FullPath"/> from a string path by converting it to an absolute path.</summary>
-    /// <param name="path">The path to convert. Can be relative or absolute.</param>
+    /// <param name="path">The path to convert. Can be relative or absolute. An empty string returns <see cref="Empty"/>.</param>
     public static FullPath FromPath(string path)
     {
+        // A path that resolves to nothing already returns Empty below; an empty input is the same case, and
+        // Path.GetFullPath would throw before reaching it. Environment.GetFolderPath returns "" for a folder
+        // that is not defined on the platform, which makes this a routine input rather than a caller error.
+        if (path is { Length: 0 })
+            return Empty;
```
meziantou, line 449: Should it return the same as CurrentDirectory?

meziantou: I think `Empty` is the right answer, and `CurrentDirectory` would actually be harmful here.

The case that motivated the change is `Environment.GetFolderPath` returning `""` for a folder that is not defined on the platform. Probing on macOS:

```
GetFolderPath(CommonDesktopDirectory) = ""
GetFolderPath(Windows)                = ""
```

`""` there means "this folder does not exist", not "the current directory". If `FromPath("")` resolved against the CWD, `FullPath.GetFolderPath(SpecialFolder.Windows)` would quietly return the process working directory on macOS — a real, writable location — and code that probes a special folder and then writes to it would write to the wrong place. Throwing (today) is bad; silently returning a valid-looking path is worse.

There is also no BCL precedent for `"" == CWD`. `.` is the spelling for the current directory; `""` is rejected:

```
Path.GetFullPath("")   throws ArgumentException
Path.GetFullPath(".")  = /private/tmp/.../probe   (== Environment.CurrentDirectory)
```

And `Empty` keeps `FromPath(p.Value) == p` intact, since `FullPath.Empty.Value` is `""`. Returning `CurrentDirectory` would break that round-trip for exactly one value.

I pushed the reasoning as a comment on the guard plus two tests: `FromPath("")` is not `CurrentDirectory()` while `FromPath(".")` is, and the round-trip holds for both `Empty` and a normal path. If you would rather `FromPath` keep throwing and only `GetFolderPath` special-case the empty result, that also solves the original problem — say the word and I will restructure it that way.

PR #2344, opened 2026-08-29
Parenthesize the expressions the FullPath code fixes generate

meziantou:
Eight code fix providers built `receiver.Member` or `a / b` with no precedence handling. The base class has had a `Parenthesize` helper for exactly this since it was written, but only two call sites used it.

The worst case **compiles and returns the wrong path**, because `FullPath` converts implicitly to `string`:

```csharp
// before
return Path.GetFileName(condition ? a : b);

// after applying the fix — parses as condition ? a : (b.Name), types as string
return condition ? a : b.Name;
```

When `condition` is true that yields the whole path instead of the file name. No compiler diagnostic, and the diff looks like the tidy-up that was intended. The same shape affects `.Extension`, `.Parent`, `.NameWithoutExtension`, `ChangeExtension` and `MakePathRelativeTo`.

The rest are build breaks:

```csharp
Path.GetFileName((FullPath)value)        → (FullPath)value.Name        // CS1061 on object.Name
Path.Combine(root, name ?? "default")    → root / name ?? "default"    // (root / name) ?? … → CS0019
Path.IsPathRooted(p).ToString()          → !p.IsEmpty.ToString()       // !(string) → CS0023
```

## What changed

- `Parenthesize` moves out of `FullPathCodeFixProviderBase` into an `ExpressionSyntaxExtensions` extension method, so the ten providers that do **not** derive from the base can reach it. Logic is unchanged.
- Every generated member-access receiver and both operands of every generated `/` now go through it.
- `FullPathCodeFixProviderBase.ApplyFixAsync` parenthesizes the replacement when the node it replaces was itself a receiver. That is what fixes the `!p.IsEmpty` case, and it applies to every provider deriving from the base rather than to that one call site.

Parenthesization is skipped for primary expressions, so `fullPath.Name` and `root / "sub"` are emitted exactly as before — all 105 pre-existing tests pass unchanged, which is the check that this does not add noise to ordinary fixes.

## Verification

Four code-fix tests added asserting the exact fixed source: conditional receiver, cast receiver, `??` operand, and predicate-as-receiver. All four fail on `main` and pass with the fix. Full analyzer suite green at 109 tests.

## Note

This is layer 1 of a 3-PR stack. It is self-contained and can merge on its own.

### DIFF
--- a/src/Meziantou.Framework.Roslyn/ExpressionSyntaxExtensions.cs
+++ b/src/Meziantou.Framework.Roslyn/ExpressionSyntaxExtensions.cs
@@ -1,17 +1,24 @@
+#if !MEZIANTOU_FRAMEWORK_ROSLYN_ENABLE_WARNINGS
+#pragma warning disable
+#endif
+#nullable enable
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
-namespace Meziantou.Framework.Analyzers.FullPath;
+namespace Meziantou.Framework.Roslyn;
 
-internal static class ExpressionSyntaxExtensions
+#if !MEZIANTOU_FRAMEWORK_ROSLYN_DISABLE_EMBEDDEDATTRIBUTE
+[Microsoft.CodeAnalysis.Embedded]
+#endif
+internal static partial class ExpressionSyntaxExtensions
 {
     /// <summary>
     /// Wraps the expression in parentheses unless it is already a primary expression.
     /// </summary>
     /// <remarks>
-    /// Generated code appends to the expression, either as the target of a member access or as an operand of the
-    /// <c>/</c> operator. Both bind tighter than most expressions, so an operand such as <c>a ? b : c</c> has to be
-    /// parenthesized to keep the meaning of the original code.
+    /// Generated code usually appends to an expression, either as the target of a member access or as an operand of a
+    /// binary operator. Both bind tighter than most expressions, so an operand such as <c>a ? b : c</c> has to be
+    /// parenthesized to preserve the meaning of the original code.
     /// </remarks>
     public static ExpressionSyntax Parenthesize(this ExpressionSyntax expression)
     {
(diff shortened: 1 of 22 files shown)


src/Meziantou.Framework.Roslyn/ExpressionSyntaxExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Wraps the expression in parentheses unless it is already a primary expression.
+    /// </summary>
+    /// <remarks>
+    /// Generated code appends to the expression, either as the target of a member access or as an operand of the
+    /// <c>/</c> operator. Both bind tighter than most expressions, so an operand such as <c>a ? b : c</c> has to be
+    /// parenthesized to keep the meaning of the original code.
+    /// </remarks>
+    public static ExpressionSyntax Parenthesize(this ExpressionSyntax expression)
```
meziantou, line 23: This should be part of the Meziantou.Framework.Roslyn package, and referenced in this project.

meziantou: Moved to `src/Meziantou.Framework.Roslyn/ExpressionSyntaxExtensions.cs`. The `<Compile Include="../Meziantou.Framework.Roslyn/**/*.cs" />` glob in the CodeFix project already picks it up, so the reference side needed nothing beyond a `using Meziantou.Framework.Roslyn;` in the ten call sites.

Two things that fell out of the move:

- It now packs into the source package (`Content Include="**/*.cs"`), so it compiles against Roslyn 4.8 through 5.6. Verified all four `Meziantou.Framework.Roslyn.Tests` variants build clean.
- `Meziantou.Framework.Roslyn.Tests` compiles with `MEZIANTOU_FRAMEWORK_ROSLYN_ENABLE_WARNINGS`, so MA0182 fired on it as an unused internal type. I added `Parenthesize` tests to `RoslynHelperTests` rather than suppressing the rule — 16 theory cases covering the primary-expression whitelist and the wrapped cases, plus idempotency.

I left `Version` at 1.0.4, since version bumps look like they happen in their own commits here (#1972, 1bf3808ef) rather than alongside changes.

PR #2400, opened 2026-08-29
Dispose the current Process after snapshotting it

meziantou:
## The finding

`CurrentProcessSnapshot.cs:9-12` — the base constructor receives `Process.GetCurrentProcess()` and reads ~18 properties from it, but nothing disposes the `Process`.

### Failure scenario

`Process` holds a `SafeProcessHandle` until finalisation. Small and bounded per call, but this is a library — a host that snapshots on every unhandled exception accumulates handles between collections. It also breaks the convention the rest of this repo follows, where every other `IDisposable` is in a `using`.

## What changed

`CurrentProcessSnapshot` now chains through a private constructor that disposes the `Process` after the base constructor has read every property.

Ownership stays where it belongs: `ProcessSnapshot(Process)` still does **not** dispose a `Process` it did not create — only `CurrentProcessSnapshot`, which creates one, disposes it. Confirmed no other caller constructs a `ProcessSnapshot`.

Field-initialiser ordering is safe here: with `this(…)` chaining, the derived initialisers run inside the private constructor before `base(…)`, and none of them touch `process`.

## Review feedback

The first revision also disposed each `ProcessModule` from `process.Modules`. That was wrong and has been reverted: on .NET Core `ProcessModule` owns no unmanaged handle — `ProcessManager` closes the module handles before returning the collection — so `Component.Dispose` is a no-op there. `ProcessSnapshot.cs` is now byte-identical to `main`.

## Verification

10/10 tests pass on net10.0 + net11.0, including `IsJsonSerializable`, which builds a full `CurrentProcessSnapshot` and serializes it — so the disposal does not race the property reads. Handle-count behaviour is Windows-specific and was not measured.

### DIFF
--- a/src/Meziantou.Framework.Diagnostics.ContextSnapshot/ProcessSnapshot.cs
+++ b/src/Meziantou.Framework.Diagnostics.ContextSnapshot/ProcessSnapshot.cs
@@ -13,7 +13,7 @@ public class ProcessSnapshot
         StartTime = process.StartTime;
 
         MainModule = process.MainModule is null ? null : new ProcessModuleSnapshot(process.MainModule);
-        Modules = process.Modules.Cast<ProcessModule>().Select(module => new ProcessModuleSnapshot(module)).ToImmutableArray();
+        Modules = GetModules(process);
 
         UserProcessorTime = process.UserProcessorTime;
         TotalProcessorTime = process.TotalProcessorTime;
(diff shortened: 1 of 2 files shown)


src/Meziantou.Framework.Diagnostics.ContextSnapshot/ProcessSnapshot.cs
```diff
@@ -13,7 +13,7 @@ internal ProcessSnapshot(Process process)
         StartTime = process.StartTime;
 
         MainModule = process.MainModule is null ? null : new ProcessModuleSnapshot(process.MainModule);
-        Modules = process.Modules.Cast<ProcessModule>().Select(module => new ProcessModuleSnapshot(module)).ToImmutableArray();
+        Modules = GetModules(process);
```
meziantou: ProcessModule doesn't need to be disposed. Revert this file.

meziantou: Reverted in 7bf5f1a — `ProcessSnapshot.cs` is now byte-identical to `main`.

You're right, and my reasoning was stale: I was carrying over the .NET Framework behaviour. On .NET Core `ProcessManager` closes the module handles before returning the collection, so `ProcessModule` owns nothing unmanaged and `Component.Dispose` is a no-op.

The PR is now just the `Process` disposal, which does own a `SafeProcessHandle`. Title and description updated to match.

PR #2412, opened 2026-08-29
Write iCalendar date-times in a form RFC 5545 defines

meziantou:
**Stack 2 of 2 · merges into `fix/scheduling-ics-escape` · merge #2411 first**

## The bug

`src/Meziantou.Framework.Scheduling/Utilities.cs:57-65`

```csharp
public const string DateTimeFormat = "yyyyMMddTHHmmsszzz";   // note the offset

public static string DateTimeToString(DateTime dt)
{
    if (dt.Kind is DateTimeKind.Utc)
        return dt.ToString(UtcDateTimeFormat, CultureInfo.InvariantCulture);

    return dt.ToString(DateTimeFormat, CultureInfo.InvariantCulture);
}
```

For any `DateTime` whose `Kind` is not `Utc` — including `Unspecified`, which is what `new DateTime(2024, 1, 2, 8, 0, 0)` and `default(DateTime)` produce — the format appends a UTC offset. Verified output:

```
DTSTART:20240102T080000-05:00
CREATED:00010101T000000-05:17
```

RFC 5545 §3.3.5 defines exactly three DATE-TIME forms: floating local time (no suffix), UTC (`Z` suffix), and local time with a `TZID` parameter. **An inline offset is none of them**, so conforming parsers reject or misread the line. It also leaked the generating machine's time zone into the file, and for `default(DateTime)` produced historical LMT offsets like `-05:17`.

## The change

| `Kind` | Before | After |
|---|---|---|
| `Utc` | `20240102T080000Z` | unchanged |
| `Unspecified` | `20240102T080000-05:00` | `20240102T080000` (floating, form 1) |
| `Local` | `20240102T080000-05:00` | converted to UTC, `…Z` (form 2) |

Also writes **`PRODID`**, which §3.6 requires in a `VCALENDAR` and which was never emitted.

## Side benefit

`Utilities.DateTimeToString` also formats `UNTIL` in the RRULE `Text` output (`DailyRecurrenceRule.cs:108` and siblings). A rule whose `EndDate` was not `Utc` therefore emitted `UNTIL=20240101T000000+01:00`, which is not valid RRULE syntax either. That is fixed by the same change. Rules parsed from a string are unaffected — `Utilities.ParseDateTime` already returns `Kind == Utc`, so every existing `Text` round-trip test still passes untouched.

## Follow-up, not in this PR

`Event.Created` / `LastModified` / `DateTimeStamp` default to `default(DateTime)`, so an event that does not set them still serialises `00010101T000000`. Valid now, but meaningless. `DTSTAMP` is REQUIRED by §3.6.1 so it cannot simply be omitted; giving those properties a sensible default deserves its own change.

## Verification

7 new tests: `PRODID` presence, each of the three `Kind` values, and a theory asserting that no `DTSTART`/`DTEND`/`CREATED`/`LAST-MODIFIED`/`DTSTAMP` value ever matches an offset form (`^\d{8}T\d{6}Z?$`).

Full project suite: **280 passed, 0 failed** (262 before, 18 new across the stack).

### DIFF
--- a/tests/Meziantou.Framework.Scheduling.Tests/InternetCalendarTests.cs
+++ b/tests/Meziantou.Framework.Scheduling.Tests/InternetCalendarTests.cs
@@ -18,6 +18,20 @@ public sealed class InternetCalendarTests
         return count;
     }
 
+    private static string GetContentLine(string ics, string name)
+    {
+        foreach (var contentLine in ics.Split("\r\n", StringSplitOptions.RemoveEmptyEntries))
+        {
+            if (contentLine.StartsWith(name + ":", StringComparison.Ordinal))
+            {
+                return contentLine;
+            }
+        }
+
+        Assert.Fail($"No '{name}' content line in:\n{ics}");
+        return null!;
+    }
+
     private static InternetCalendar CreateCalendarWithEvent(Event @event)
     {
         var calendar = new InternetCalendar();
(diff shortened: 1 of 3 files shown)


tests/Meziantou.Framework.Scheduling.Tests/InternetCalendarTests.cs
```diff
@@ -18,6 +18,20 @@ private static int CountContentLines(string ics, string line)
...
+        foreach (var contentLine in ics.Split("\r\n", StringSplitOptions.RemoveEmptyEntries))
+        {
+            if (contentLine.StartsWith(name + ":", StringComparison.Ordinal))
+            {
+                return contentLine;
+            }
+        }
+
+        Assert.Fail($"No '{name}' content line in:\n{ics}");
+        return null!;
```
meziantou:
```suggestion
        return null;
```

meziantou:
```suggestion
        throw new UnreachableException();
```

tests/Meziantou.Framework.Scheduling.Tests/InternetCalendarTests.cs
```diff
@@ -18,6 +18,20 @@ private static int CountContentLines(string ics, string line)
...
+        foreach (var contentLine in ics.Split("\r\n", StringSplitOptions.RemoveEmptyEntries))
+        {
+            if (contentLine.StartsWith(name + ":", StringComparison.Ordinal))
+            {
+                return contentLine;
+            }
+        }
+
+        Assert.Fail($"No '{name}' content line in:\n{ics}");
+        throw new UnreachableException();
```
meziantou:
```suggestion
        throw new System.Diagnostics.UnreachableException();
```

PR #2448, opened 2026-08-29
Match the rest of the path when a pattern ends with the recursive wildcard

meziantou:
A pattern whose last segment is `**` matches nothing at all, in every dialect that supports the recursive wildcard:

```csharp
Glob.Parse("src/**", GlobDialect.Standard).IsMatch("src/a.txt")    // false
Glob.Parse("src/**", GlobDialect.Standard).IsMatch("src/x/a.txt")  // false
Glob.Parse("**",     GlobDialect.Standard).IsMatch("a.txt")        // false
Glob.Parse("foo/**", GlobDialect.Git).IsMatch("foo/a.txt")         // false
```

`IsPartialMatch` still returns `true` for those directories, so `EnumerateFiles("src/**")` walks the whole tree and then rejects every entry — slow *and* empty.

## Cause

`Glob.IsMatchCore` bails out as soon as a `RecursiveMatchAllSegment` has nothing after it:

```csharp
var remainingPatternSegments = patternSegments[(i + 1)..];
if (remainingPatternSegments.IsEmpty) // Last segment
    return false; // Match only files
```

The comment says "match only files", but the effect is "match nothing" — there is no path for which this returns `true`.

## Change

A trailing `**` now consumes the rest of the path, requiring at least one segment, so `src/**` matches the content of `src` but not `src` itself — the same rule git (`foo/**`) and MSBuild use. Every consumed segment goes through the existing `CanMatchLeadingDot` check, so `src/**` still does not reach `src/.hidden/a.txt` unless `GlobOptions.MatchLeadingDot` is set.

This is deliberately not done as a parser rewrite to `MatchNonEmptyTextSegment`: that segment skips the per-segment leading-dot checks and is only sound when `MatchLeadingDot` is on, which is why `CreateGlob` gates the existing `**` optimisations behind it. Handling it in the matcher keeps both modes correct.

Directory matching is unchanged: `src/**` still carries `GlobMatchType.File`, so directory entries are rejected before reaching this code.

## Verification

- `dotnet test tests/Meziantou.Framework.Globbing.Tests` — 378 passed (364 before, 14 added here).
- Differential fuzz against an independent reference matcher, 3,535 generated patterns x 183 generated paths, Release build: the trailing-`**` family of wrong answers — 522 pairs across the 7 pattern shapes `**`, `*/**`, `?/**`, `[a-c]/**`, `a/**`, `[ab]/**`, `[!a]/**` — **goes to 0**. Remaining fuzz failures on this branch belong to two unrelated defects, each fixed in its own PR.
- Tests cover all three dialects, the "must consume at least one segment" rule, the leading-dot interaction in both modes, and an `EnumerateFiles` case.
- `dotnet run ./eng/update-all.cs` — no changes.

One thing worth a second opinion: whether `src/**` should *also* match the directory `src` itself. I kept it not matching, to agree with git and MSBuild.

### DIFF
--- a/src/Meziantou.Framework.Globbing/Glob.cs
+++ b/src/Meziantou.Framework.Globbing/Glob.cs
@@ -208,8 +208,23 @@ public sealed class Glob : IGlobEvaluatable
             if (patternSegment is RecursiveMatchAllSegment)
             {
                 var remainingPatternSegments = patternSegments[(i + 1)..];
-                if (remainingPatternSegments.IsEmpty) // Last segment
-                    return false; // Match only files
+                if (remainingPatternSegments.IsEmpty)
+                {
+                    // A trailing '**' matches the rest of the path. It must consume at least one segment, so 'a/**'
+                    // matches the content of 'a' but not 'a' itself.
+                    if (pathReader.IsEndOfPath)
+                        return false;
+
+                    while (!pathReader.IsEndOfPath)
+                    {
+                        if (!CanMatchLeadingDot(pathReader, matchLeadingDot[i]))
+                            return false;
+
+                        pathReader.ConsumeSegment();
+                    }
+
+                    return true;
+                }
 
                 var remainingAnchors = anchors[(i + 1)..];
                 var remainingMatchLeadingDot = matchLeadingDot[(i + 1)..];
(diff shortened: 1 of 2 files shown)


meziantou:
Rebased onto `main` to resolve the conflict with #2445.

The conflict was in `GlobTests.cs` only — both PRs appended test methods at the same anchor, and git interleaved the two blocks. `Glob.cs` applied cleanly. Resolved by taking `main`'s version of the test file and re-applying this PR's five test methods, so #2445's tests are intact and this PR's diff is still just its own change.

Re-verified with #2445 now in `main`:

- `dotnet test` — **397 passed** (383 on `main`, 14 from this PR).
- Differential fuzz, 646,905 pattern/path pairs, Release: **0 crashes, 96 wrong answers** — down from 25,057 crashes and 618 wrong answers before either fix. All 96 remaining belong to the multi-wildcard backtracking defect fixed in #2453, none to this one.
- `dotnet run ./eng/update-all.cs` — no changes.

PR #2457, opened 2026-08-29
Resolve gitignore paths against the last matching pattern

meziantou:
> **Stack 1 of 2** · merges into `main` · **#2462 targets this branch and must merge after it**

git resolves a path against the **last** pattern that matches it. `GlobCollection.IsMatch` instead returns `false` as soon as any exclude matches, whatever the order:

```csharp
// .gitignore:  *.log / !important.log / important.log
GlobCollection.ParseGitIgnore(content).IsMatch("important.log");   // false - git ignores it
```

Re-including then re-excluding is a normal gitignore idiom (`!*.log` followed by `debug.log`), and the disagreement with git is silent.

The existing `LoadGitIgnore_ParsesPatterns` test only exercises the ordering where the two rules happen to agree, which is why this was not caught.

## Change

Collections built by `ParseGitIgnore` / `LoadGitIgnoreAsync` now use last-match-wins. Hand-built collections are **unchanged** and keep "any exclude wins".

That split is deliberate. "Any exclude wins" is order-independent, which is a good property for a general-purpose include/exclude collection and is what existing callers of the public constructor rely on. git's rule is the contract only where the method name promises git, so it is scoped to those two factories via a private constructor flag rather than changed globally.

## Scope: what this does not fix

git also refuses to let a negation re-include a file whose *parent directory* is excluded — `!bin/keep.txt` has no effect if `bin/` is ignored, because git never descends into `bin`. That rule needs ancestor-aware evaluation and is not implemented here; this PR is only about the ordering of matches on a single path.

## Verification

- `dotnet test tests/Meziantou.Framework.Globbing.Tests` — 368 passed (364 before, 4 added here).
- `GitIgnoreResolvesAPathAgainstTheLastMatchingPattern` covers the re-ignore case, `GitIgnoreNegationAfterAMatchReIncludesThePath` covers the plain negation, and `GitIgnoreOrderMatters` pins that the two orderings now differ.
- `HandBuiltCollectionKeepsAnyExcludeWins` pins the unchanged behaviour of the public constructor, so the split cannot be collapsed by accident.
- `dotnet run ./eng/update-all.cs` — no changes.

### DIFF
--- a/src/Meziantou.Framework.Globbing/GlobCollection.cs
+++ b/src/Meziantou.Framework.Globbing/GlobCollection.cs
@@ -187,16 +187,15 @@ public sealed class GlobCollection : IReadOnlyList<IGlobEvaluatable>, IGlobEvalu
 
     private bool IsLastMatch(ReadOnlySpan<char> directory, ReadOnlySpan<char> filename, PathItemType? itemType)
     {
-        var match = false;
-        foreach (var glob in _globs)
+        // The last pattern that matches decides, so walk backwards and stop at the first hit.
+        for (var i = _globs.Length - 1; i >= 0; i--)
         {
+            var glob = _globs[i];
             if (glob.IsMatch(directory, filename, itemType))
-            {
-                match = glob.Mode is GlobMode.Include;
-            }
+                return glob.Mode is GlobMode.Include;
         }
 
-        return match;
+        return false;
     }
 
     /// <summary>Determines whether a directory should be recursed into when enumerating files.</summary>
--- a/src/Meziantou.Framework.Globbing/GlobCollection.cs
+++ b/src/Meziantou.Framework.Globbing/GlobCollection.cs
@@ -23,9 +23,19 @@ public sealed class GlobCollection : IReadOnlyList<IGlobEvaluatable>, IGlobEvalu
 {
     private readonly IGlobEvaluatable[] _globs;
 
+    // gitignore resolves a path against the last pattern that matches it, whereas a hand-built collection uses
+    // the order-independent "any exclude wins" rule. Only the gitignore factories set this.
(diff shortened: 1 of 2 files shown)


meziantou:
Good question — I checked rather than assumed, and reordering cannot work, but half the idea is right and I have applied it.

**Why reordering cannot work.** The existing loop computes exactly `anyInclude && !anyExclude`. The `continue` and the early `return false` are short-circuits that do not change the result, so the value depends only on *which* globs match a path, never on their order. Permuting the list is a no-op.

Verified on `main` with the patterns `*.log`, `!important.log`, `important.log`, `!*.tmp`, `a*.log`, `!ab*.log` over 7 paths, trying **all 720 orderings**: every path yields exactly one distinct result across all of them. `important.log` comes out `False` in all 720, where git says ignored. No permutation reaches git's answer.

It is not just this example. git can alternate include/exclude over nested sets — `*.log` / `!a*.log` / `ab*.log` / `!abc*.log` — and `anyInclude && !anyExclude` applies excludes globally, so it can never re-include `ab*.log` after `a*.log` was excluded. Expressing that would need patterns like "`a*.log` but not `ab*.log`", which this dialect cannot write.

**What the idea does buy.** Last-match-wins is naturally "iterate backwards, first match decides", so I have rewritten `IsLastMatch` as a reverse loop with an early return instead of a forward scan that kept overwriting a variable. Same semantics, less work, and it reads much closer to how git is actually specified. Thanks — that is a real improvement.

**What is still open.** Selecting between the two semantics still needs something, since one type now has two behaviours. The alternative to the `_lastMatchWins` field is a distinct type for the gitignore case, which is cleaner but a public API addition. I went with the field to keep this PR small. Happy to switch if you would rather have the separate type.

PR #2511, opened 2026-08-29
Hold the session handle open for the duration of each native call

meziantou:
Follow-up to a project review of `Meziantou.Framework.Win32.RestartManager`.

**Stack 2 of 3 · merges into `fix/registerfiles-null-guard` (#2510) · must merge after #2510 and before #2512.**

## Finding

Every native call read the session out of the `SafeHandle` as a plain `uint` (`RestartManagerSessionHandle.cs:24`, consumed at 7 call sites in `RestartManager.cs`). That opts out of the protection `SafeHandle` exists to provide, in two ways:

1. `ObjectDisposedException.ThrowIf(_sessionHandle.IsClosed, this)` is a time-of-check/time-of-use test — nothing stops a `Dispose()` between the check and the P/Invoke.
2. Once `_sessionHandle.SessionHandle` has been read, neither the `RestartManager` nor its `RestartManagerSessionHandle` is reachable for the rest of the method, so the GC may collect the wrapper and run `SafeHandle`'s critical finalizer — calling `RmEndSession` on a session a native call is still using.

When a `SafeHandle` is passed to a P/Invoke *as a `SafeHandle`*, the marshaller does `DangerousAddRef`/`DangerousRelease` around the call and neither can happen. The Restart Manager takes the session as a `DWORD`, so that never applied here.

## What I could not show

I tried to reproduce both variants and **failed**, and this PR should not claim otherwise. The reachable calls are too fast to hit: `RegisterFiles(8000)` completes in 7 ms and `RegisterFiles(50000)` in 28 ms, and a forced `GC.Collect()`/`WaitForPendingFinalizers()` loop over 200 abandon-and-collect iterations produced zero failures.

The exposure is concentrated in the call I could not safely test: `RmShutdown` blocks for up to 30 seconds per unresponsive application (30 s for apps, 20 s for services), and a GC inside a 30-second window under load is close to certain. It also needs the caller to have dropped their reference, which the documented `using var session = …` pattern prevents — the risk shape is `var s = CreateSession(); …; s.Shutdown(…);` with no `using`.

So: a real contract violation with a narrow window today and a wide one on the least-tested path, not a demonstrated failure. Reviewers may reasonably weigh it differently, which is why it sits above the null-guard fix rather than below it.

## What changed

- `RestartManagerSessionHandle.Acquire()` returns a `Scope` that takes a `DangerousAddRef` and releases it on `Dispose`, holding a reference for the whole call.
- The raw `SessionHandle` property is **removed**, so the scope is now the only way to obtain the handle and the old pattern cannot creep back.
- All 7 call sites (`RegisterFile`, `RegisterFiles`, `IsResourcesLocked`, `GetList`, `Shutdown`, `Restart`, `CancelCurrentTask`) go through `using var session = _sessionHandle.Acquire();`.
- `GetList` acquires once around its whole retry loop rather than per iteration.

The existing `ObjectDisposedException.ThrowIf` checks are kept: `DangerousAddRef` would also throw on a closed handle, but the explicit check keeps the clearer message and the existing test expectations.

`Scope` is a `readonly struct`, so this adds no allocation.

## Tests

No test can pin the race itself. Added `Operations_AfterDispose_ThrowObjectDisposedException`, which exercises the new `Acquire()` path on a disposed session across `RegisterFiles`, `GetProcessesLockingResources`, `GetLockingProcesses`, `Shutdown`, `Restart` and `CancelCurrentTask` — five of which had no coverage at all. It is safe because the disposed check fires before any native call.

## Verification

- Tests on this branch: `total: 15, failed: 0, succeeded: 15` — 14 from #2510 plus the one added here.
- No public API change, so `ref/` is unchanged.

### DIFF
--- a/src/Meziantou.Framework.Win32.RestartManager/RestartManagerSessionHandle.cs
+++ b/src/Meziantou.Framework.Win32.RestartManager/RestartManagerSessionHandle.cs
@@ -7,8 +7,17 @@ namespace Meziantou.Framework.Win32;
 
 /// <summary>Owns a Restart Manager session handle and ends the session when released.</summary>
 /// <remarks>
+/// <para>
 /// The Restart Manager documents no invalid session handle value, and 0 is a session like any other,
 /// so -1 is used as the "no handle" sentinel instead of the usual 0.
+/// </para>
+/// <para>
+/// The Restart Manager takes the session as a <see cref="uint"/> rather than as a handle, so its value has to be read
+/// out of this <see cref="SafeHandle"/>. Reading it directly would leave nothing keeping this instance reachable for
+/// the rest of the call, so the garbage collector could run the finalizer and end the session while a call is still
+/// using it. Callers must go through <see cref="SafeHandleValue"/>, which holds a reference for the duration of the
+/// call and also prevents a concurrent <see cref="SafeHandle.Dispose()"/> from ending the session mid-call.
+/// </para>
 /// </remarks>
 [SupportedOSPlatform("windows6.0.6000")]
 internal sealed class RestartManagerSessionHandle : SafeHandle
@@ -20,8 +29,6 @@ internal sealed class RestartManagerSessionHandle : SafeHandle
     }
 
     public override bool IsInvalid => handle == -1;
-
-    public uint SessionHandle => (uint)handle;
 
     protected override bool ReleaseHandle()
     {
--- a/src/Meziantou.Framework.Win32.RestartManager/Meziantou.Framework.Win32.RestartManager.csproj
+++ b/src/Meziantou.Framework.Win32.RestartManager/Meziantou.Framework.Win32.RestartManager.csproj
@@ -9,6 +9,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="../Meziantou.Framework/SafeHandleValue.cs" Link="SafeHandleValue.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.3.298">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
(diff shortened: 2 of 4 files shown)


src/Meziantou.Framework.Win32.RestartManager/RestartManagerSessionHandle.cs
```diff
@@ -21,10 +21,44 @@ public RestartManagerSessionHandle(uint sessionHandle)
...
-    public uint SessionHandle => (uint)handle;
+    /// <summary>Holds the session open for the duration of a native call and exposes the raw handle to pass to it.</summary>
+    /// <remarks>
+    /// The Restart Manager takes the session as a <see cref="uint"/> rather than as a handle, so its value has to be read
+    /// out of this <see cref="SafeHandle"/>. Reading it directly would leave nothing keeping this instance reachable for
+    /// the rest of the call, so the garbage collector could run the finalizer and end the session while a call is still
+    /// using it. Going through this scope takes a reference that prevents that, and also prevents a concurrent
+    /// <see cref="SafeHandle.Dispose()"/> from ending the session mid-call.
+    /// </remarks>
+    public Scope Acquire()
```
meziantou: Reuse https://github.com/meziantou/Meziantou.Framework/blob/main/src/Meziantou.Framework/SafeHandleExtensions.cs instead of a new implementation

meziantou: Done in e2dca6f — dropped the `Acquire()`/`Scope` implementation and reused `SafeHandleValue`.

Followed the same pattern as `Meziantou.Framework.Win32.Jobs`: source-linked the file rather than taking a project reference, so the package keeps its zero-dependency footprint.

```xml
<Compile Include="../Meziantou.Framework/SafeHandleValue.cs" Link="SafeHandleValue.cs" />
```

All 7 call sites are now `using var handleScope = new SafeHandleValue(_sessionHandle);` with `(uint)handleScope.Value`, matching `JobObject`. `RestartManagerSessionHandle` is back to a minimal diff against `main` — the only code change left is removing the raw `SessionHandle` property, so the unsafe pattern can't come back. I kept the rationale as a `<remarks>` on the class since that context isn't obvious from the call sites.

Unlike Jobs this didn't need the `UnscopedRefAttribute` polyfill exclusion — it builds clean without it.

The layer diff drops from +63/−8 to +41/−9. Builds clean for `net10.0`/`net11.0` with `/p:TreatWarningsAsErrors=true`, 15/15 tests pass, and `ref/` is unchanged since `SafeHandleValue` stays internal here. #2512 was rebased on top.

PR #2542, opened 2026-08-29
Report an unknown zone instead of Restricted for paths at MAX_PATH

meziantou:
> **Stack 5 of 8** · merges into `fix/motw-exception-contracts` (#2524)

## The problem

`MapUrlToZone` stops reading the file once the path reaches `MAX_PATH` and answers `URLZONE_UNTRUSTED`. The zone reported for a longer path was therefore never read from the file at all. Measured, on freshly created local files:

```
len= 258 zone=Internet       (marked)
len= 259 zone=Internet       (marked)
len= 260 zone=Untrusted      <== boundary, exactly MAX_PATH
len= 261 zone=Untrusted
```

Both directions are wrong: an **unmarked** local file at 284 characters came back `Untrusted` / `IsUntrusted == true`, and a file genuinely marked `Internet` came back `Untrusted` (4) rather than `Internet` (3). .NET enables long paths on modern Windows, so callers legitimately hold such paths.

## What I tried first, and why it isn't in this PR

My review suggested normalizing the path. Both candidate forms were implemented and measured, and both are worse:

| Approach | Result |
|---|---|
| `\\?\` extended-length prefix | urlmon rejects it outright — `ArgumentException` (`E_INVALIDARG`) out of `MapUrlToZone`. `IsUntrusted` has no catch, so it crashes. |
| `file://` URL via `new Uri(path).AbsoluteUri` | Does **not** fix long paths — still `Untrusted` at 294 characters. And it regresses non-ASCII names: a marked `café.txt` that correctly reads `Internet` today reads back `LocalMachine` / `untrusted=false`. A silently *lost* mark is the worst possible direction. |

urlmon simply cannot resolve a zone for a path at or beyond `MAX_PATH`.

## What changed

`GetFileZone` returns `UrlZone.Invalid` when the full path is 260 characters or more — which is what its contract already documents for a zone it cannot determine — rather than a zone it fabricated.

`IsUntrusted` returns `true` for the same paths. That is the existing behavior, now deliberate and documented: it matches the fail-closed choice the method already makes when the security manager cannot be created. The false positive on unmarked long-path files remains, and is the honest answer while the zone is unreadable — fixing it properly would mean parsing `Zone.Identifier` directly and reimplementing the saved-file check.

## Verification

`dotnet test -p:TreatWarningsAsErrors=true` — 58/58 pass on `net10.0` and `net11.0`.

New tests pin all three sides of the boundary: `GetFileZone` returns `Invalid` at exactly 260, `IsUntrusted` returns `true` at 260, and both still resolve normally at 259.

### DIFF
--- a/src/Meziantou.Framework.Win32.MarkOfTheWeb/MarkOfTheWeb.cs
+++ b/src/Meziantou.Framework.Win32.MarkOfTheWeb/MarkOfTheWeb.cs
@@ -46,7 +46,7 @@ public static class MarkOfTheWeb
 
     /// <summary>Gets the security zone of a file using the Windows Security Manager COM API.</summary>
     /// <param name="filePath">The path to the file to query.</param>
-    /// <returns>The <see cref="UrlZone"/> of the file, or <see cref="UrlZone.Invalid"/> if the zone cannot be determined.</returns>
+    /// <returns>The <see cref="UrlZone"/> of the file, or <see cref="UrlZone.Invalid"/> if the zone cannot be determined, which includes a full path of 260 characters or more.</returns>
     [SupportedOSPlatform("windows")]
     public static UrlZone GetFileZone(string filePath)
     {
@@ -56,6 +56,8 @@ public static class MarkOfTheWeb
             return UrlZone.Invalid;
 
         filePath = Path.GetFullPath(filePath);
+        if (filePath.Length >= MaxPathLength)
+            return UrlZone.Invalid;
 
         try
         {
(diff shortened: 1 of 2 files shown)


meziantou:
Closing this one. The length check encoded a Windows/urlmon limitation inside the library, which is the wrong place for it — `CoInternetCreateSecurityManager` should be left to report whatever it reports, rather than having the library second-guess it with a hard-coded `MAX_PATH` constant.

The underlying observation still stands (`MapUrlToZone` stops resolving at exactly 260 characters and answers `URLZONE_UNTRUSTED`, even for an unmarked local file whose `Zone.Identifier` .NET reads without trouble at 400 characters), but that is Windows' behaviour to own, not this library's to paper over.

The commit has been dropped from the stack; #2543, #2544 and #2546 are rebased onto #2524 and no longer reference it.

PR #2687, opened 2026-08-30
Keep partial multi-byte sequences buffered in the output decoder

meziantou:
**Stack 1 of 2 · merges into `main` · must merge before the shared-sink fix.**

## Finding

`StringBuilderOutputTarget.Write` and `LineBasedTextOutputTarget.Write` returned early when `GetCharCount` reported zero characters. `Decoder.GetCharCount` is a pure query — it does **not** consume the bytes; only `GetChars` stores an incomplete trailing sequence in the decoder's state. So when a read returned *only* the leading bytes of a multi-byte character, those bytes were thrown away and the next read decoded an orphaned continuation byte as `U+FFFD`.

Reproduced against a real process emitting `é` as two writes (`printf '\303'; sleep 0.3; printf '\251'`):

```
stringbuilder = <?>  (hex EFBFBD0A)   expected 'é' (C3A9…)
lines = [<?>]                          expected [é]
CORRUPTED: True
```

Every text-capturing path is affected, including the default `ExecuteBufferedAsync` — `LineBasedTextOutputTarget` is the base for `ToTextDelegate`, `ToTextWriter` and `ToProcessOutputCollection`. The read buffer is 4096 bytes and pipe reads return whatever is available, so any child that flushes mid-character silently loses one. It fails as data corruption, not an exception, so callers get no signal.

## Change

Always hand the bytes to `GetChars` so the decoder keeps the incomplete sequence for the next call, and append only when it actually produced characters.

## Verification

Two new tests drive a `FakeProcess` over a stream that returns one byte per read, making the split deterministic without a real process — one for `ExecuteBufferedAsync`, one for the `StringBuilder` target. Both fail on `main` and pass here.

Full ProcessWrapper suite on net10.0 + net11.0: 183/183 passing.


tests/Meziantou.Framework.ProcessWrapper.Tests/ProcessWrapperTests.cs
```diff
@@ -2013,5 +2045,47 @@ public void Intercept(ProcessWrapper processWrapper, ProcessStartInfo processSta
         }
     }
 
+    /// <summary>A stream that returns a single byte per read, so multi-byte characters are split across reads.</summary>
+    private sealed class SingleByteReadStream(byte[] bytes) : Stream
```
meziantou: Use RestrictedStream.cs

meziantou: Done in 5833debc3 — `RestrictedStream` with `MaxReadLength = 1`, which also applies the limit to `ReadAsync(Memory<byte>)`, the overload the output pump uses. Dropped `SingleByteReadStream` (-29 lines in the test file).

The test project now references `Meziantou.Framework`, so `update-all.cs` added it to `Meziantou.Framework.slnx`.

Re-checked that the tests still catch the bug: both fail on the parent commit and pass with the fix.

PR #2690, opened 2026-08-30
Do not search the working directory for the executable by default

meziantou:
## Finding

`ResolveFileName` passed the working directory to `ExecutableFinder.GetFullExecutablePath`, which searches it **before** `PATH`. A bare command name therefore resolved to a file sitting in the working directory.

Reproduced with a shell script named `git` dropped into a temp directory:

```
--- hijack ---
output: PWNED   (real git would print 'git version ...')
```

That is `ProcessWrapper.Create("git").WithArguments("--version").WithWorkingDirectory(dir)` running `dir/git`, not `/usr/bin/git`.

Windows searches the current directory by design, so the behaviour is defensible there. No POSIX shell does, precisely because of this attack — and the natural use of this library is exactly the exposed shape: `ProcessWrapper.Create("git" | "npm" | "dotnet").WithWorkingDirectory(checkoutPath)`. A CI runner or bot that clones untrusted code and then runs a tool inside the clone executes attacker-controlled code from a file the attacker only had to commit and mark executable.

## Change

The working directory is searched only when the caller opts in with `WithSearchWorkingDirectory()`.

`ExecutableFinder` itself is untouched — the `workingDirectory` parameter keeps its current contract, so `Meziantou.Framework`, `Meziantou.Framework.SnapshotTesting`, `Meziantou.Framework.InlineSnapshotTesting` and `Meziantou.Framework.TemporaryContainers` are unaffected, and `ExecutableFinderTests` still passes unchanged. Only the default used by `ProcessWrapper` changes.

**This is a breaking change** for callers that relied on resolving an executable out of the working directory. They add `WithSearchWorkingDirectory()`.

## Verification

New test `WithWorkingDirectory_DoesNotFindExecutableInWorkingDirectoryByDefault` puts an executable in the working directory and asserts it is *not* used. It fails on `main` (the script runs, no exception) and passes here.

Three existing tests relied on the old default and now opt in explicitly: `WithWorkingDirectory_FindsExecutableInWorkingDirectory`, `StartAndForget_AppliesConfigurationAndInterceptors`, and the first via the resolution path. They still assert the same behaviour, now behind the opt-in.

`ref/Meziantou.Framework.ProcessWrapper.cs` regenerated for the new method; a Release build with `/p:PublicApiGeneratorVerifyNoChangeOnBuild=true` is clean. `readme.md` documents the default and the opt-in.

Full ProcessWrapper suite on net10.0 + net11.0: 181/181 passing.

### DIFF
--- a/src/Meziantou.Framework.ProcessWrapper/ProcessWrapper.cs
+++ b/src/Meziantou.Framework.ProcessWrapper/ProcessWrapper.cs
@@ -228,9 +228,16 @@ public sealed class ProcessWrapper
 
     /// <summary>Also looks for the executable in the working directory, before the directories listed in <c>PATH</c>.</summary>
     /// <remarks>
-    /// Disabled by default. When the executable is a bare file name, enabling this lets a file in the working
-    /// directory take precedence over the command found in <c>PATH</c>, so only enable it when the working
-    /// directory is trusted.
+    /// <para>
+    /// Windows resolves a bare executable name against the current directory before <c>PATH</c>, so this is the
+    /// native behavior there. POSIX shells deliberately do not.
+    /// </para>
+    /// <para>
+    /// It is nevertheless disabled by default on every platform, Windows included, for security reasons: a file
+    /// sitting in the working directory would otherwise take precedence over the command <c>PATH</c> resolves for
+    /// the same name, letting an untrusted working directory substitute its own executable for the intended one.
+    /// Only enable it when the working directory is trusted.
+    /// </para>
     /// </remarks>
     public ProcessWrapper WithSearchWorkingDirectory(bool searchWorkingDirectory = true)
     {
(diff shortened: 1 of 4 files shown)


src/Meziantou.Framework.ProcessWrapper/ProcessWrapper.cs
```diff
@@ -224,6 +226,18 @@ public ProcessWrapper WithWorkingDirectory(string workingDirectory)
         return this;
     }
 
+    /// <summary>Also looks for the executable in the working directory, before the directories listed in <c>PATH</c>.</summary>
```
meziantou, line 229: Mention that this is the default behavior on Windows, but disabled by default for security reasons.

meziantou: Done in b77d01bb0. The remarks now say that resolving a bare name against the current directory is the native Windows behavior, that POSIX shells deliberately do not do it, and that it is disabled by default on every platform — Windows included — because an untrusted working directory could otherwise substitute its own executable.

Added the same note to `readme.md`, since the Windows default is the part most likely to surprise someone reading only the docs.

PR #2698, opened 2026-08-30
Throw instead of silently discarding sections left open by an HTML email template

meziantou:
## Finding

`HtmlEmailOutput.BeginSection` pushes onto `_currentSections`; only `EndSection` moves the captured
text into `_sections` (`HtmlEmailOutput.cs:126-150`). A template that opens a section and never
closes it renders its body into the main output normally, but the section never reaches the metadata:

```csharp
var template = new HtmlEmailTemplate();
template.Load("Hello {{@begin_section title}}Subject");

var result = template.Run(out var metadata);
// result:         "Hello Subject"   ✔
// metadata.Title: null              ✘  — silently
```

For the `title` section that means an email goes out with an empty subject line, and nothing anywhere
reports a problem.

## Change

An unbalanced section is a broken template, so it is now rejected rather than papered over:

```
TemplateException: The template is invalid: no end_section directive for 'title'.
TemplateException: The template is invalid: no end_section directive for 'title', 'footer'.
```

The check lives in an `InvokeRunMethod` override rather than in `GetMetadata`, so **every** `Run`
overload validates — including the inherited ones that never produce an `HtmlEmailMetadata`.
`ThrowIfSectionsAreOpen` is `internal`: `HtmlEmailTemplate` is the only caller, and keeping it off
the public surface avoids a `ref/` API change.

`TemplateException` is the type the readme documents for template errors, so this stays inside the
existing error contract.

**This is a breaking change** for templates that today render with an unclosed section and quietly
lose the metadata. That is the intent — see the discussion below.

## Verification

Four tests; three fail on `main`:

- ✘ an unclosed section throws, and the message names both `end_section` and `'title'`,
- ✘ several unclosed sections are all reported,
- ✘ the inherited `Template.Run()` path throws too,
- ✔ same-named nested sections still keep the outermost content — a guard, passing before and after.

`Meziantou.Framework.Templating.Html.Tests` 26/26, `Meziantou.Framework.Templating.Tests` 114/114,
`Meziantou.Framework.Templating.Tool.Tests` 22/22, on net10.0 + net11.0. Built with
`/p:TreatWarningsAsErrors=true`; `dotnet run ./eng/update-all.cs` reports no changes.

## Still silent: the mirror case

`EndSection` with no matching open section (`{{@end_section}}` on its own) does nothing at all —
`_currentSections.LastOrDefault(...)` returns null and the method falls through. Same class of
invalid template, equally silent. Left out of this PR to keep it to what was reviewed; happy to fold
it in here or file it separately.


src/Meziantou.Framework.Templating.Html/HtmlEmailOutput.cs
```diff
@@ -149,6 +149,22 @@ public void EndSection(string name)
         }
     }
 
+    /// <summary>
```
meziantou: I think it should throw and report the invalid template, instead of silently close sections.

meziantou: Agreed — changed in dde7502.

An unclosed section now throws `TemplateException` rather than being closed behind the author's back:

```
The template is invalid: no end_section directive for 'title'.
The template is invalid: no end_section directive for 'title', 'footer'.
```

Two things I did beyond a straight swap:

- The check moved out of `GetMetadata` into an `InvokeRunMethod` override, so every `Run` overload validates — including the inherited `Template.Run()` ones that never build an `HtmlEmailMetadata`. Validation shouldn't depend on whether the caller asked for metadata.
- All open sections are reported in one message, not just the first.

Tests updated to match: unclosed throws, multiple unclosed are all named, and the inherited-overload path throws too. The nested same-name guard test is unchanged.

One thing I left alone: the mirror case, `{{@end_section}}` with nothing open, is still silent — `EndSection` finds no match and falls through. Same class of invalid template. Want me to fold that into this PR, or keep it separate?

PR #2706, opened 2026-08-30
Stop the pathname canonicalization from inserting a separator before every literal

meziantou:
`CanonicalizePathname` is the encoding callback that `PatternParser` runs on **every** fixed-text part of a parsed pattern — plus each part's prefix and suffix — not once on the whole pathname. It prepended `/` to any value that did not already start with one, so any literal that follows a group acquired a separator:

| pattern | compiled `Pathname` (before) | `IsMatch` (before) |
|---|---|---|
| `/books/:id.json` | `/books/:id/.json` | `https://e.com/books/123.json` → **false** |
| `/assets/*.png` | `/assets/*/.png` | `https://e.com/assets/logo.png` → **false** |
| `/:name-suffix` | `/:name/-suffix` | `https://e.com/value-suffix` → **false** |
| `/file-:name-v:version.txt` | `/file-:name/-v:version/.txt` | `https://e.com/file-report-v2.txt` → **false** |

Callers got a silent `false` from `IsMatch` — no exception, no diagnostic — so a route table or allow-list built on an extension pattern quietly matched nothing. The `Pathname` getter reported the mangled pattern back, so the one value that would have revealed the problem was itself wrong.

Patterns whose parts all begin with `/` (`/a/:x/b`) were unaffected, which is why this went unnoticed.

## What changed

`CanonicalizePathname` is now the identity. The spec's [canonicalize a pathname](https://urlpattern.spec.whatwg.org/#canonicalize-a-pathname) does percent-encoding only — it deliberately prefixes `"/-"` before running the URL parser and strips those two code points back off afterwards, precisely so that a value which does not start with `/` round-trips unchanged. This implementation does no percent-encoding, so the faithful callback is a no-op.

The one behaviour this drops is an undocumented convenience: `new UrlPatternInit { Pathname = "books" }` previously became `/books`. It now stays `books`, which is what the spec and every browser implementation do. No existing test relied on it.

## Verification

- 11 new theory/fact cases in `UrlPatternTests` covering an extension after a group, a wildcard with an extension, a literal suffix, several groups with literals between them, and the constructor-string path. Each of them fails on `main`.
- `Match_ExtensionAfterAGroup_ShouldCaptureWithoutTheExtension` also asserts the group captures `123` (not `123.json`) and that the previously-matching `/books/123/.json` no longer matches.
- Full suite: 550 passed, 0 failed, both target frameworks, no warnings.
- `dotnet run ./eng/update-all.cs` produces no changes.

### DIFF
--- a/src/Meziantou.Framework.Uri/UrlPattern/UrlPattern.cs
+++ b/src/Meziantou.Framework.Uri/UrlPattern/UrlPattern.cs
@@ -769,15 +769,10 @@ public sealed class UrlPattern
 
     private static string CanonicalizePathname(string value)
     {
-        if (string.IsNullOrEmpty(value))
-            return value;
-
-        // Ensure starts with /
-        if (!value.StartsWith('/', StringComparison.Ordinal))
-        {
-            value = "/" + value;
-        }
-
+        // The callback runs on every fixed-text part of the pattern, not once on the whole pathname,
+        // so it must not add a leading "/": that would insert a separator in front of any literal
+        // that follows a group, turning "/books/:id.json" into "/books/:id/.json".
+        // The spec avoids the same trap by prefixing "/-" before parsing and stripping it afterwards.
         return value;
     }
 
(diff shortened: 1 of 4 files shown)


src/Meziantou.Framework.Uri/UrlPattern/UrlPattern.cs
```diff
@@ -769,15 +769,10 @@ private static string CanonicalizePort(string value)
 
     private static string CanonicalizePathname(string value)
```
meziantou: It seems the method is useless now

PR #2712, opened 2026-08-30
Insert the target framework into a shared embedded constants output path

meziantou:
A multi-targeted project whose `EmbeddedConstantsOutputPath` does not vary by target framework has its inner builds race on one generated file.

## The finding

`build/Meziantou.Framework.EmbeddedConstantsGenerator.targets:16-18` resolves `_EmbeddedConstantsGeneratedCodePath` from `EmbeddedConstantsOutputPath`. The default is under `$(IntermediateOutputPath)`, which is target-framework-specific, so the default is safe.

But the readme documents `EmbeddedConstantsOutputPath` as a user-settable property with no mention of multi-targeting. Set it to something target-framework-independent — `Generated/EmbeddedConstants.g.cs`, the natural choice if you want the file visible in the IDE — on a project with `<TargetFrameworks>net10.0;net11.0</TargetFrameworks>`, and both inner builds resolve to the same path and run `GenerateEmbeddedConstants` concurrently.

`WriteFileIfChanged` (`GenerateEmbeddedConstantsTask.cs:76-88`) does `File.ReadAllText` then `File.WriteAllText` with no lock, no share mode and no retry, so the two nodes can collide on a sharing violation, or one can read while the other writes.

## The change

Make the path correct instead of rejecting the configuration. When the project is multi-targeted and the resolved output path does not already contain the target framework, the target framework is inserted into it: `Generated/EmbeddedConstants.g.cs` becomes `Generated/net10.0/EmbeddedConstants.g.cs`. The default path already varies and is left alone.

### Keeping the generated files out of the default glob

Once the generated files sit inside the project directory, the SDK's default `**/*.cs` glob picks up *every* framework's copy, and the build fails with `CS0102: the type 'EmbeddedConstants' already contains a definition for 'SampleText'`. The first build passes only because the files do not exist yet at evaluation time; the second one fails. This was reachable before this change too, for anyone who wrote `Generated/$(TargetFramework)/...` by hand.

So the targets file also removes the generated files from `Compile` at evaluation time, and the target adds this framework's file back. Two MSBuild details drove the shape of that, and are commented in the file:

- The removal has to happen at evaluation time, not inside the target — a wildcard `Remove` there did not match.
- `Compile Remove="$(SomeProperty)"` does **not** glob. MSBuild escapes wildcards that arrive through property expansion, so the pattern is matched literally and hits nothing; `$([MSBuild]::Unescape(...))` restores globbing. Confirmed directly: the same pattern as a literal expanded to 2 items while the property form stayed at 1 unexpanded item.

## Verification

Two tests, each building **twice**, because the second build is where the glob problem bites:

- `GenerateOnBuild_MultiTargetingWithSharedOutputPath_InsertsTheTargetFrameworkIntoThePath` — `net10.0;net11.0` with `Generated/EmbeddedConstants.g.cs` builds twice and produces `Generated/net10.0/…` and `Generated/net11.0/…`, and nothing at `Generated/EmbeddedConstants.g.cs`.
- `GenerateOnBuild_MultiTargetingWithPerTargetFrameworkOutputPath_GeneratesOneFilePerTargetFramework` — the same project with `Generated/$(TargetFramework)/EmbeddedConstants.g.cs` builds twice and produces two generated files.

Checked both are not vacuous: with the targets change reverted, both fail, the second with the `CS0102` above. Full suite 8/8 on net10.0 and net11.0.

Readme updated to describe the insertion.

## Note for reviewers

An earlier revision of this PR rejected the configuration with a new `MFECG0010` diagnostic. That was replaced per review feedback, so `MFECG0010` is now unused — the other PRs from this review took `MFECG0011`, `MFECG0012` and `MFECG0013`, leaving a gap. Say the word if you would rather they were renumbered down.


src/Meziantou.Framework.EmbeddedConstantsGenerator/build/Meziantou.Framework.EmbeddedConstantsGenerator.targets
```diff
@@ -26,6 +26,12 @@
   <Target Name="GenerateEmbeddedConstants"
           BeforeTargets="CoreCompile;BeforeCompile"
           Condition="'@(EmbeddedConstant)' != '' and '$(IsCrossTargetingBuild)' != 'true'">
+    <!-- The inner builds of a multi-targeted project run in parallel. If they all resolve to the same
```
meziantou: Not the right fix. Instead, update the `Meziantou_EmbeddedConstantsOutputPath` so it contains the TFM

meziantou: Done in caf6b8602 — the error is gone and the target framework is now inserted into `Meziantou_EmbeddedConstantsOutputPath` when the project is multi-targeted and the path does not already contain it. `Generated/EmbeddedConstants.g.cs` becomes `Generated/net10.0/EmbeddedConstants.g.cs`; the default path under `IntermediateOutputPath` already varies and is left alone.

One thing worth flagging, because it made this bigger than a straight swap. Once the generated files sit inside the project directory, the SDK's default `**/*.cs` glob picks up *every* framework's copy, so the second build fails with `CS0102: the type 'EmbeddedConstants' already contains a definition for 'SampleText'`. The first build passes only because the files do not exist yet at evaluation time. That also affected a path you wrote yourself as `Generated/$(TargetFramework)/...`, so it was reachable before this change too.

So the targets file now also removes the generated files from `Compile` at evaluation time and lets the target add this framework's file back. Two MSBuild details that cost me a few rounds and are the reason for the comments in the file:

- The removal has to happen at evaluation time, not inside the target — `Remove` with a wildcard did not match there.
- `Compile Remove="$(SomeProperty)"` does **not** glob: MSBuild escapes wildcards that arrive through property expansion, so the pattern is matched literally and hits nothing. `$([MSBuild]::Unescape(...))` restores it. I confirmed this directly — the same pattern written as a literal expanded to 2 items while the property form stayed at 1 unexpanded item.

Both multi-targeting tests now build twice, since the second build is where this actually breaks. Verified they fail without the targets change (with the `CS0102` above) and pass with it; full suite is 8/8 on net10.0 and net11.0.

This frees up `MFECG0010`. The other PRs from this review took 0011, 0012 and 0013, so there is now a gap — happy to renumber them down if you would rather keep the sequence contiguous.

PR #2728, opened 2026-08-30
Pick the wildcard tool directory by name instead of last write time

meziantou:
`EnumerateDirectories` expanded a `*` segment with
`.OrderByDescending(Directory.GetLastWriteTimeUtc)`, i.e. most recently *modified*, which
is not the same as newest *version*. A directory's timestamp changes whenever its direct
children change, so an update, a repair install, or any stray write under an older
version promotes it above a newer one.

Reproduced with a synthetic tree where only the oldest entry was touched last:

```
Microsoft Visual Studio/{2019,2022,2026}/Community/Common7/IDE/
-> picked 2019
```

This affects every wildcard search directory in the table: `VisualStudio`
(`Microsoft Visual Studio\*\…`), `BeyondCompare` (`Beyond Compare *`), `Rider`
(`Rider*`, `JetBrains Rider *`) and `Vim` (`vim*`).

A developer with VS 2022 and VS 2026 side by side gets a choice that can silently change
between runs. Both accept `/diff`, so it is an annoyance rather than a break — but "why
did it open the old one this time" is hard to diagnose from the outside.

## What changed

Order wildcard matches by directory name using
`CompareOptions.NumericOrdering | CompareOptions.IgnoreCase` (invariant culture). That is
deterministic, and numeric-aware so variable-width version numbers rank correctly:

|  | numeric | ordinal |
|---|---|---|
| `vim9`, `vim10` | `vim10` > `vim9` | `vim9` > `vim10` |
| `Beyond Compare {4,5,10}` | `10` > `5` > `4` | `5` > `4` > `10` |
| `Rider 2025.{1,2,10}` | `2025.10` > `2025.2` > `2025.1` | `2025.2` > `2025.10` > `2025.1` |
| `{2019,2022,2026}` | `2026` > `2022` > `2019` | same |

An earlier revision of this PR used `StringComparer.OrdinalIgnoreCase`, which is correct
for every pattern in `Definitions` *today* but ranks `Beyond Compare 10` below
`Beyond Compare 4` — reachable rather than theoretical, since Beyond Compare is on v5.
Switched to numeric ordering after review feedback.

`CompareOptions.NumericOrdering` is .NET 10+; both target frameworks are net10.0 and
net11.0, so no conditional compilation is needed.

## Verification

Two tests, each confirmed to fail against the ordering it guards against and pass with
this change:

- `TryFindByName_WildcardDirectoryPrefersTheHighestVersion` builds `{2019,2022,2026}` and
  touches `2019` last — fails under last-write-time ordering (`Expected: "2026" / Actual: "2019"`).
- `TryFindByName_WildcardDirectoryOrdersVersionsNumerically` builds `{vim9,vim10}` — fails
  under ordinal ordering (`Expected: "vim10" / Actual: "vim9"`).

`dotnet test /p:TreatsWarningsAsErrors=true` — 12 of 14 pass on net10.0 and net11.0. The
two failures are `TryFindByExtension_ReturnsTextTool`, pre-existing and unrelated: it
depends on which diff tools the developer has installed. Fixed in a sibling PR; CI images
are bare, so this branch is green there.


meziantou:
P2: ordinal string sorting does not select the highest version once a version component has different widths. For example, both `Rider 2025.9` and `Rider 2025.10` match the built-in Rider wildcard, but descending ordinal order puts `Rider 2025.9` first, so the older installation is selected. The test's fixed-width years mask this. Use a version-aware/natural comparison for versioned tool directories, or avoid calling this the newest version if directory-name ordering is the intended behavior.

src/Meziantou.Framework.DiffEngine/DiffTools.cs
```diff
@@ -418,10 +418,13 @@ private static IEnumerable<string> EnumerateDirectories(string directoryPattern)
...
                     try
                     {
+                        // Order by name so the newest version wins deterministically. Last-write time looks like a
+                        // proxy for "newest" but a directory's timestamp changes whenever its direct children do,
+                        // so an update to an older version would promote it above a newer one.
                         nextRoots.AddRange(
                             Directory
                                 .EnumerateDirectories(root, segment)
-                                .OrderByDescending(Directory.GetLastWriteTimeUtc));
+                                .OrderByDescending(Path.GetFileName, StringComparer.OrdinalIgnoreCase));
```
meziantou: Would it be possible to use an numeric sort (CompareOptions.NumericOrdering)

meziantou: Yes — done in 8c9fefd.

```c#
private static readonly StringComparer DirectoryNameComparer =
    StringComparer.Create(CultureInfo.InvariantCulture, CompareOptions.NumericOrdering | CompareOptions.IgnoreCase);
```

It fixes more than the `vim9`/`vim10` case I'd flagged. Ordinal put `Beyond Compare 10` **last**, below `Beyond Compare 4`:

|  | numeric | ordinal |
|---|---|---|
| `vim9`, `vim10` | `vim10` > `vim9` | `vim9` > `vim10` |
| `Beyond Compare {4,5,10}` | `10` > `5` > `4` | `5` > `4` > `10` |
| `Rider 2025.{1,2,10}` | `2025.10` > `2025.2` > `2025.1` | `2025.2` > `2025.10` > `2025.1` |
| `{2019,2022,2026}` | unchanged | unchanged |

Beyond Compare is on v5 today, so that one was reachable rather than theoretical.

`CompareOptions.NumericOrdering` is .NET 10+, and both target frameworks are net10.0/net11.0, so no conditional compilation needed. Added `TryFindByName_WildcardDirectoryOrdersVersionsNumerically`, confirmed it fails against the ordinal comparer (`Expected: "vim10" / Actual: "vim9"`) and passes with this change. Existing tests unaffected — 12 of 14 pass, the two failures being the unrelated pre-existing `TryFindByExtension_ReturnsTextTool`.

PR #2737, opened 2026-08-30
Verify a dummy password hash when the Identity user does not exist

meziantou:
## Finding

`HttpBasicAuthenticationIdentityExtensions.cs:77-79` returned immediately when `FindByNameAsync` yielded `null`, skipping password verification entirely. A valid username with a wrong password ran the full PBKDF2 verification; an unknown username did not.

Measured over 30 requests each, same host, after warmup:

```
known user   + wrong password : 41.843 ms/req
unknown user + wrong password :  0.139 ms/req
ratio                         : 302.0x
```

A 302x gap is not a subtle side channel — it is trivially measurable across the internet through normal jitter, and it lets an attacker enumerate valid accounts before spending any effort on passwords.

In fairness this mirrors ASP.NET Core Identity's own `SignInManager.PasswordSignInAsync`, so it is an inherited pattern rather than a novel mistake — but a Basic-auth front door is a much more exposed place to inherit it, and the fix is local to this file.

## What changed

When the user lookup misses, the handler now verifies the supplied password against a cached dummy hash before returning `null`, so both paths do comparable work.

The dummy hash is produced once by a standalone `PasswordHasher<object>`. That deliberately avoids `IPasswordHasher<TUser>`: `VerifyHashedPassword` requires a `TUser` instance, and `TUser` is constrained only to `class` — there is no `new()` constraint to construct one, and `Activator.CreateInstance<TUser>()` would be both failure-prone and hostile to this package's `IsTrimmable` setting. Using a separate hasher keeps `TUser` out of it entirely and cannot break a custom `IPasswordHasher<TUser>` implementation.

The tradeoff, noted in a comment: the dummy hash uses Identity's **default** `PasswordHasherOptions`. An application that customises the iteration count will see the two paths cost approximately, not exactly, the same. That removes the dominant signal (300x) and leaves noise.

No public API change, so `ref/` is untouched.

## Verification

`AspNetCoreIdentity_UnknownUser_DoesNotShortCircuitPasswordVerification` measures both paths and asserts the ratio. Confirmed it **fails against the unfixed code**, reproducing the finding almost exactly:

```
failed  AspNetCoreIdentity_UnknownUser_DoesNotShortCircuitPasswordVerification
  Actual:   335.2928837311145
  total: 10  failed: 1  succeeded: 9
```

**On the timing assertion:** it compares two paths that, after the fix, do the same work, so the ratio is insensitive to how loaded the machine is — a slow runner slows both sides equally. The bound is 10x against a pre-fix value of 335x, so the margin is large. If you would still rather not carry a timing-sensitive test in CI, drop that test and keep `AspNetCoreIdentity_UnknownUser_IsRejected`, which is a plain functional check.

Full suite green on both target frameworks with `/p:TreatsWarningsAsErrors=true`: **20 passed** (10 x net10.0/net11.0). `dotnet run ./eng/update-all.cs` reports no changes.

### DIFF
--- a/src/Meziantou.AspNetCore.Authentication.HttpBasic/HttpBasicAuthenticationIdentityExtensions.cs
+++ b/src/Meziantou.AspNetCore.Authentication.HttpBasic/HttpBasicAuthenticationIdentityExtensions.cs
@@ -9,6 +9,13 @@ namespace Meziantou.AspNetCore.Authentication.HttpBasic;
 /// <summary>Extension methods to register HTTP Basic authentication with ASP.NET Core Identity.</summary>
 public static class HttpBasicAuthenticationIdentityExtensions
 {
+    // Verifying this hash when the user does not exist keeps the presence of an account from being
+    // observable through response time. It is produced once, with Identity's default options; if the
+    // application customises PasswordHasherOptions the two paths cost approximately, not exactly, the same.
+    private static readonly PasswordHasher<object> DummyPasswordHasher = new();
+    private static readonly object DummyPasswordHasherUser = new();
+    private static readonly string DummyPasswordHash = DummyPasswordHasher.HashPassword(DummyPasswordHasherUser, "8OqA2Zs4kkGm1sJv");
+
     /// <summary>Adds HTTP Basic authentication using ASP.NET Core Identity to validate credentials and build user principals.</summary>
     public static AuthenticationBuilder AddHttpBasicIdentity<TUser>(this AuthenticationBuilder builder)
         where TUser : class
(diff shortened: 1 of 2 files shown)


meziantou:
P1: the dummy verification is always performed with a new default `PasswordHasher<object>`, not the hasher configured for `UserManager<TUser>`. If the application raises `PasswordHasherOptions.IterationCount` (or replaces `IPasswordHasher<TUser>`), a known account performs the configured, potentially much more expensive verification while an unknown account performs the default 100,000-iteration PBKDF2 verification. That preserves a measurable username-enumeration oracle—the case this change is intended to remove. Generate and verify the dummy hash through the application's configured password hasher (and add a test with a non-default hasher/options).

PR #2739, opened 2026-08-30
Canonicalize a UrlPatternInit used as a match input

meziantou:
`Match(UrlPatternInit)` and `IsMatch(UrlPatternInit)` ran `ProcessUrlPatternInit` over the input — which only resolves the base URL — and then handed the raw component strings to the compiled regexes. The *pattern* side is canonicalized when it is compiled, so the two sides were compared in different normal forms:

```csharp
var p = UrlPattern.Create(new UrlPatternInit { Protocol = "https", Hostname = "example.com" });

p.IsMatch(new UrlPatternInit { Protocol = "HTTPS",  Hostname = "example.com" });  // false
p.IsMatch(new UrlPatternInit { Protocol = "https:", Hostname = "EXAMPLE.com" });  // false
```

The spec runs *process a URLPatternInit* with type `"url"` over the input, which lowercases the protocol, strips a single trailing `:`, and strips the leading `?`/`#` from the search and hash. Without that, every caller building a `UrlPatternInit` from parsed URL parts has to know to normalize first, and nothing said so.

## What changed

A `CanonicalizeInitComponents` helper applies the existing canonicalization callbacks — `CanonicalizeProtocol`, `CanonicalizeHostname`, `CanonicalizeSearch`, `CanonicalizeHash` — to an init used as a match input. `MatchInit` and `IsMatchInit` both go through it. The `Uri` and `string` overloads are untouched: `System.Uri` already hands over canonical components, which is why `MatchUrl` does not go through this path.

`UrlPatternResult.Inputs` keeps reporting the init the caller supplied rather than the canonicalized form, so `Match` still tells you what you passed in. That is what `MatchComponents`'s new `originalInit` parameter is for.

Deliberately **not** included, to keep this to one defect each:

- The pathname is left alone. `CanonicalizePathname` currently prepends a `/`, which is wrong for a per-part callback and is being fixed separately; the spec's version does no percent-encoding-free work anyway, so applying it here would be a no-op once that lands.
- The port is left alone, because `CanonicalizePort` is currently a no-op — also a separate change.
- The username and password are left alone. There is a related inconsistency there worth its own look: the pattern side escapes literals with `Uri.EscapeDataString`, while `MatchUrl` runs `Uri.UnescapeDataString` over `Uri.UserInfo`, so a pattern of `Username = "a b"` compiles to `a%20b` and is matched against `a b`. That is a separate defect from this one and is not touched here.

## Verification

Nine new cases in `UrlPatternTests`; six fail on `main`:

- `IsMatch_InitInput_CanonicalizesTheProtocol` — three theory cases (`HTTPS`, `https:`, `HTTPS:`).
- `IsMatch_InitInput_CanonicalizesTheHostname`
- `IsMatch_InitInput_StripsTheSearchAndHashPrefixes`
- `Match_InitInput_CanonicalizesTheInputAndReportsItUnchanged` — also asserts `Inputs` still carries the supplied values.

`IsMatch_InitInput_StillRejectsAValueThatDoesNotMatch` pins that canonicalizing did not make matching permissive.

Full suite: 542 passed, 0 failed, both target frameworks, no warnings. `dotnet run ./eng/update-all.cs` produces no changes.


meziantou:
P1: `CanonicalizeInitComponents` only canonicalizes protocol, hostname, search, and hash. The pattern side canonicalizes username, password, port, and pathname too, so identical init dictionaries can now fail to match. For example, a pattern with `Username = "a b"` compiles that component as `a%20b`, but `IsMatch(new UrlPatternInit { Username = "a b" })` supplies the unescaped string. Please run all input components through the corresponding URL-input canonicalization (including special-vs-opaque pathname handling).

PR #2754, opened 2026-08-30
Normalize a numeric port instead of leaving CanonicalizePort a no-op

meziantou:
`CanonicalizePort` looped over the value checking `char.IsDigit`, then returned `value` from inside the loop and again after it — both arms identical, so the whole method was a no-op with a dead branch.

The visible consequence is that a port written with leading zeros never collapses to the scheme's default, so it matches nothing:

```csharp
var p = UrlPattern.Create(new UrlPatternInit { Protocol = "https", Port = "0443" });
p.Port;                            // "0443" — expected ""
p.IsMatch("https://example.com/");  // false  — expected true
```

`https://example.com/` reports no port (it is the default), so the pattern's literal `0443` can never match it. `Port = "08080"` had the same problem against `:8080`.

## What changed

`CanonicalizePort` now parses a value made only of ASCII digits and re-serializes it, so `0443` → `443` and `08080` → `8080`. Anything that is not a plain number — `*`, `:p`, `80{80}?` — is returned untouched and matched as written, and so is a number that does not fit a port (`99999`), since it cannot be a real port anyway.

`Create` also normalizes `processedInit.Port` before comparing it against the scheme's default port, which is where the collapse to `""` happens; without that, normalizing only inside the per-part callback would come too late.

Deliberately not included: the spec throws a `TypeError` for a port that is neither a number nor a valid pattern. Doing that here would turn currently-accepted input into an exception in a shipped package, so this keeps accepting it. Happy to follow up if you would rather it throw.

## Verification

Eight new cases in `UrlPatternTests`:

- `Create_PortWithLeadingZeros_CollapsesLikeTheDefaultPort` and `Create_PortWithLeadingZeros_MatchesTheSamePort` — fail on `main`.
- `Create_NumericPort_IsNormalized` — `08080` → `8080`, and `8080`/`0` unchanged.
- `Create_PortThatIsNotAPlainNumber_IsMatchedAsWritten` — `*`, `99999` and `:p` pass through, so wildcard and group ports still work.

Full suite: 544 passed, 0 failed, both target frameworks, no warnings. `dotnet run ./eng/update-all.cs` produces no changes.


meziantou:
P1: this normalizes a pattern port but init inputs are still matched verbatim. After this change, `UrlPattern.Create(new UrlPatternInit { Port = "08080" }).IsMatch(new UrlPatternInit { Port = "08080" })` returns false: the pattern is `8080` while the input remains `08080`. Canonicalize the port for `UrlPatternInit` match inputs as well (or otherwise preserve equivalent matching).

PR #2758, opened 2026-08-30
Invalidate node caches with a document version instead of walking ancestors

meziantou:
Loading a deeply nested document is quadratic in its depth: a 352 KB fragment of nested `<div>`s takes **30 seconds** of CPU.

### The failure

`HtmlNodeList.OnCollectionChanged` calls `ClearCaches()` on the parent for every node added, and `ClearCaches` walked from that node up to the document, nulling five cached strings on each ancestor. Parsing adds one node per level of nesting, so the walk costs `O(depth)` per node and `O(n²)` overall.

Measured with `new HtmlDocument().LoadHtml("<div>" * N + "x" + "</div>" * N)`, Release, arm64:

| N | input | before | after |
|---|---|---|---|
| 8 000 | 88 KB | 567 ms | 116 ms |
| 16 000 | 176 KB | 3 447 ms | 54 ms |
| 32 000 | 352 KB | 29 872 ms | 205 ms |
| 64 000 | 704 KB | — | 288 ms |

The walk was pure waste during parsing: nothing reads a cached value while the document is being built, so every field it nulled was already `null`.

This matters most for `Meziantou.Framework.HtmlSanitizer`, which is built on this parser and exists to process untrusted HTML — a request body far below any default size limit pins a core for half a minute.

### The change

Invalidation moves from "clear every ancestor" to "move one counter". `HtmlDocument` carries a `CacheVersion`; `ClearCaches` increments it; each node stamps its cached values with the version they were computed for and drops them when the stamp no longer matches. Invalidation is now `O(1)` and covers the whole document rather than just one ancestor chain.

One case needed care. `OwnerDocument` is not maintained for the descendants of a removed node, so after `p.Remove()` the descendants of `p` still report their changes against the document while `p` itself no longer sees them. A node with no `OwnerDocument` therefore keeps no cached value at all — it recomputes on every read. That preserves the existing behaviour exactly (a first attempt that only compared versions returned a stale `p.OuterHtml`, which `HtmlNode_CachedValuesAreRecomputedOnADetachedSubtree` now pins down). Detached roots lose their caching; they are rare, and correctness wins.

`ClearCaches` keeps its signature and its meaning, so an override still works.

### Verification

Five tests in `HtmlNodeTests`:

- `HtmlNode_LoadHtml_DeeplyNestedDocumentIsNotQuadratic` reproduces the defect — 20 000 levels against a 5 s budget. On `main` it fails at **19 s**; it now needs about a tenth of a second, so the budget has roughly 50x of headroom in both directions.
- Four cache-invalidation tests (descendant text change, descendant removal, attribute change, detached subtree). These pass on `main` too — they are there to pin the behaviour this change had to preserve, and one of them caught a real regression in an earlier attempt.

Green after the change:

| project | tests |
|---|---|
| `Meziantou.Framework.Html.Tests` | 280 |
| `Meziantou.Framework.HtmlSanitizer.Tests` | 336 |
| `Meziantou.Framework.HtmlToMarkdown.Tests` | 1698 |
| `Meziantou.Framework.Html.Tool.Tests` | 12 |
| `Meziantou.Framework.Templating.Html.Tests` | 18 |

No public API change, and `dotnet run ./eng/update-all.cs` reports no further changes.

Found during a review of `Meziantou.Framework.HtmlSanitizer`.


meziantou:
P1: cache invalidation uses the changed node's `OwnerDocument`, which can be stale after moving a detached subtree between documents. Detach `<p><span>a</span></p>` from document A, append it to document B, cache B's HTML, then change the span: the span still points to A, so this increments A's `CacheVersion` and B serves stale cached HTML. The ownership/version needs to follow the current ancestor chain (or ownership must propagate through the subtree when it is reparented).

PR #2769, opened 2026-08-30
Escape paths when building diff tool arguments

meziantou:
Every argument builder was `$"\"{temp}\" \"{target}\""`-shaped, wrapping paths in quotes
with no escaping. A quote is a legal file name character on Linux and macOS, so a path
containing one ends the argument early and the rest becomes extra arguments to the diff
tool:

```c#
tool.GetArguments("/tmp/a\" --wait --extensionDevelopmentPath=/evil \"x.txt", "/tmp/verified.txt")
// --diff "/tmp/a" --wait --extensionDevelopmentPath=/evil "x.txt" "/tmp/verified.txt"
```

`--wait` and `--extensionDevelopmentPath=/evil` are now real arguments to VS Code. Both
snapshot-testing packages pass this string straight to
`new ProcessStartInfo(resolvedTool.ExePath, arguments)`.

The realistic failure is corruption rather than attack — a snapshot file whose name came
from a parameterised test case containing a quote produces a garbled command line, a diff
of the wrong files, or an opaque launch failure. But the flags reachable this way
(`--extensionDevelopmentPath`, `--install-extension`) are ones you would rather not have
reachable at all.

A trailing backslash has the same problem from the other end: `"C:\dir\"` escapes its own
closing quote.

## What changed

Escaping is delegated to the repo's existing `CommandLineBuilder`, imported with
`<Compile Include>` the same way `ProcessWrapper` and `TemporaryContainers` do — no project
reference:

```xml
<Compile Include="../Meziantou.Framework.CommandLine/CommandLineBuilder.cs" Link="Utils/CommandLineBuilder.cs" />
```

Each path now goes through `CommandLineBuilder.WindowsQuotedArgument`, which implements the
rules `ProcessStartInfo.Arguments` is parsed with on every platform, not just Windows.

**This changes the generated command line for ordinary paths.** `WindowsQuotedArgument`
only quotes when the value contains one of `' '`, `\t`, `\n`, `\v`, `"`:

```
before:  --diff "/tmp/received.txt" "/tmp/verified.txt"
after:   --diff /tmp/received.txt /tmp/verified.txt
```

Both parse to the same `argv`. `GetArguments_HonorsTargetPosition`, which predates this PR,
is updated accordingly, and the argument-builder theory now uses paths containing a space so
it still exercises and displays the quoting branch — the WinMerge and Visual Studio rows show
both at once, with quoted paths and unquoted `/dl` / `/dr` titles.

### Includes the argument-builder coverage from the closed #2768

This PR previously sat on top of #2768 in a stack. Since that PR is now closed and its
work was never merged, its commit is carried here — the escaping tests depend on the
`internal static DiffTools.GetLaunchArguments(DiffTool)` hook it introduced, without which
`WinMerge` and `VisualStudio` cannot be reached from a test at all (they have no `Osx` or
`Linux` platform settings, so they never resolve on a non-Windows machine).

That commit adds a table-driven theory over all 25 tools × both `DiffEngine_TargetOnLeft`
values. Before it, only VS Code had any coverage — 10 of the 11 argument builders were
untested, including the two that compute display titles via `Path.GetFileName`. The theory
rows also pin the tool → builder mapping, which was previously only visible by reading the
200-line `Definitions` table: `Cursor` shares `VisualStudioCode`'s builder and `Neovim`
shares `Vim`'s, and nothing would have caught one being pointed at the wrong one.

**If you closed #2768 because you did not want that table**, say so and I will strip it
back to just the `GetLaunchArguments` hook plus the two escaping tests. I kept it on the
assumption that closing the lower PR was about collapsing the stack rather than rejecting
the tests.

## Verification

Beyond the unit tests, I round-tripped the output through .NET's real argument parser —
feeding the escaped string to `ProcessStartInfo.Arguments` on a process that prints its
`argv`, and checking the two original paths come back intact:

```
PASS  /tmp/received.txt /tmp/verified.txt
PASS  "/tmp/my snapshots/received.txt" "/tmp/my snapshots/verified.txt"
PASS  "/tmp/a\" --wait --extensionDevelopmentPath=/evil \"x.txt" /tmp/verified.txt
PASS  "C:\my dir\\" "C:\other dir\\"
PASS  C:\dir\ C:\other\
PASS  "/tmp/quote\".txt" /tmp/back\slash.txt
PASS  "/tmp/tab\there.txt" /tmp/b.txt
```

Note the two trailing-backslash cases: quoted, the backslash is doubled; unquoted, it is left
alone — both correct.

`dotnet test /p:TreatsWarningsAsErrors=true` — 64 of 66 pass on net10.0 and net11.0,
rebased onto current `main`. The two failures are the pre-existing, unrelated
`TryFindByExtension_ReturnsTextTool`, fixed by #2736.

## Not in this PR

Adding a `GetArgumentList` returning `IReadOnlyList<string>` so consumers could use
`ProcessStartInfo.ArgumentList` and skip string quoting entirely. That is the more robust
design, but it is a public API addition, and `ArgumentList` is mutually exclusive with the
`UseShellExecute = true` both `DiffEngineTool` implementations set — so it needs a decision
about shell-execute semantics rather than a mechanical change. Escaping correctly fixes the
defect on its own.

---

**Merge note:** this PR and #2745 touch adjacent lines of `DiffTools.cs` — #2745 deletes
`NormalizeExtension`, which sits directly above the argument builders rewritten here. They
are logically independent and can merge in either order, but git reports an adjacent-hunk
conflict on whichever goes second (re-verified after this rebase). Resolution is to keep
both sides.

### DIFF
--- a/src/Meziantou.Framework.DiffEngine/DiffTools.cs
+++ b/src/Meziantou.Framework.DiffEngine/DiffTools.cs
@@ -238,6 +238,17 @@ public static class DiffTools
         return resolvedTool is not null;
     }
 
+    internal static LaunchArguments GetLaunchArguments(DiffTool tool)
+    {
+        foreach (var definition in Definitions)
+        {
+            if (definition.Tool == tool)
+                return definition.LaunchArguments;
+        }
+
+        throw new ArgumentOutOfRangeException(nameof(tool), tool, message: null);
+    }
+
     private static List<ResolvedTool> ResolveAvailableTools()
     {
         var result = new List<ResolvedTool>();
--- a/src/Meziantou.Framework.DiffEngine/DiffTools.cs
+++ b/src/Meziantou.Framework.DiffEngine/DiffTools.cs
@@ -550,47 +550,10 @@ public static class DiffTools
     /// into extra arguments for the diff tool.
     /// </summary>
     /// <remarks>
-    /// Escaping follows the rules <see cref="System.Diagnostics.ProcessStartInfo.Arguments"/> is parsed with on
-    /// every platform, described in
-    /// <see href="https://learn.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way">Everyone quotes command line arguments the wrong way</see>.
-    /// The value is always quoted, so paths without special characters keep the command line they had before.
+    /// <see cref="CommandLineBuilder"/> implements the rules <see cref="System.Diagnostics.ProcessStartInfo.Arguments"/>
+    /// is parsed with on every platform, and only adds quotes when the value actually needs them.
     /// </remarks>
-    private static string Quote(string path)
-    {
-        var sb = new StringBuilder(path.Length + 2);
-        sb.Append('"');
-
-        for (var i = 0; i < path.Length; i++)
-        {
-            var backslashes = 0;
-            while (i < path.Length && path[i] == '\\')
-            {
-                i++;
-                backslashes++;
-            }
-
-            if (i == path.Length)
-            {
-                // Backslashes that would otherwise escape the closing quote.
-                sb.Append('\\', backslashes * 2);
-                break;
-            }
-
-            if (path[i] == '"')
-            {
-                sb.Append('\\', (backslashes * 2) + 1);
-            }
-            else
-            {
-                sb.Append('\\', backslashes);
-            }
-
-            sb.Append(path[i]);
-        }
-
-        sb.Append('"');
-        return sb.ToString();
-    }
+    private static string Quote(string path) => CommandLineBuilder.WindowsQuotedArgument(path);
 
     private static PlatformSettings Windows(string[] executableNames, params string[] searchDirectories) => new(executableNames, searchDirectories);
     private static PlatformSettings Linux(string[] executableNames, params string[] searchDirectories) => new(executableNames, searchDirectories);
--- a/src/Meziantou.Framework.DiffEngine/DiffTools.cs
+++ b/src/Meziantou.Framework.DiffEngine/DiffTools.cs
@@ -504,44 +504,92 @@ public static class DiffTools
         return trimmed[0] == '.' ? trimmed : "." + trimmed;
     }
 
-    private static string StandardLeftArguments(string temp, string target) => $"\"{target}\" \"{temp}\"";
-    private static string StandardRightArguments(string temp, string target) => $"\"{temp}\" \"{target}\"";
+    private static string StandardLeftArguments(string temp, string target) => $"{Quote(target)} {Quote(temp)}";
+    private static string StandardRightArguments(string temp, string target) => $"{Quote(temp)} {Quote(target)}";
 
-    private static string RiderLeftArguments(string temp, string target) => $"diff \"{target}\" \"{temp}\"";
-    private static string RiderRightArguments(string temp, string target) => $"diff \"{temp}\" \"{target}\"";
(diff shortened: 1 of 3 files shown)


src/Meziantou.Framework.DiffEngine/DiffTools.cs
```diff
@@ -493,44 +504,92 @@ private static string NormalizeExtension(string extension)
...
+    /// name on Linux and macOS, and an unescaped one would end the argument early and turn the rest of the path
+    /// into extra arguments for the diff tool.
+    /// </summary>
+    /// <remarks>
+    /// Escaping follows the rules <see cref="System.Diagnostics.ProcessStartInfo.Arguments"/> is parsed with on
+    /// every platform, described in
+    /// <see href="https://learn.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way">Everyone quotes command line arguments the wrong way</see>.
+    /// The value is always quoted, so paths without special characters keep the command line they had before.
+    /// </remarks>
+    private static string Quote(string path)
```
meziantou: Can you use the code from CommandLine project? (Import the file only, do not add a reference to the project)

meziantou: Done in 75c0e66 — imported via `<Compile Include>` like `ProcessWrapper` and `TemporaryContainers` do, no project reference:

```xml
<Compile Include="../Meziantou.Framework.CommandLine/CommandLineBuilder.cs" Link="Utils/CommandLineBuilder.cs" />
```

The ~35-line local escaper is gone, replaced by `private static string Quote(string path) => CommandLineBuilder.WindowsQuotedArgument(path);` — net -32 lines in `DiffTools.cs`.

**One behavior change worth flagging before you merge.** `WindowsQuotedArgument` returns the value *unquoted* when it contains none of `' '`, `\t`, `\n`, `\v`, `"`, so the generated command line loses its quotes for ordinary paths:

```
before:  --diff "/tmp/received.txt" "/tmp/verified.txt"
after:   --diff /tmp/received.txt /tmp/verified.txt
```

Both parse to the same `argv`, and this is why I'd originally hand-rolled an always-quoting version. Since you asked for the shared code, I took the churn: `GetArguments_HonorsTargetPosition` (which predates this PR) and the 25 theory rows are updated. The theory now uses paths containing a space so the table still exercises and shows the quoting branch, while `GetArguments_HonorsTargetPosition` covers the unquoted one. The WinMerge and Visual Studio rows show both at once — quoted paths, unquoted `/dl` / `/dr` titles.

Re-ran the round-trip check against .NET's real parser (escaped string → `ProcessStartInfo.Arguments` → a process that prints its `argv`):

```
PASS  /tmp/received.txt /tmp/verified.txt
PASS  "/tmp/my snapshots/received.txt" "/tmp/my snapshots/verified.txt"
PASS  "/tmp/a\" --wait --extensionDevelopmentPath=/evil \"x.txt" /tmp/verified.txt
PASS  "C:\my dir\\" "C:\other dir\\"
PASS  C:\dir\ C:\other\
PASS  "/tmp/quote\".txt" /tmp/back\slash.txt
PASS  "/tmp/tab\there.txt" /tmp/b.txt
```

Note the two trailing-backslash cases: quoted, the backslash is doubled; unquoted, it is left alone — both correct, and the second is a case my own version never produced. I changed `GetArguments_EscapesBackslashesBeforeTheClosingQuote` to use `C:\my dir\` so it still exercises the doubling.

64 of 66 tests pass on net10.0 and net11.0; the two failures are the unrelated pre-existing `TryFindByExtension_ReturnsTextTool` that #2736 fixes.

PR #2778, opened 2026-08-30
Make byte members mirror the file for every kind

meziantou:
**Stack 3 of 4** · merges into #2777 · must merge after #2776 and #2777, before #2779

The same file produces different bytes depending on its `Kind`.

## The finding

`EmbeddedConstantsGeneratorTask.cs:578` and `:588` — the text path passes `textBytes` (byte order mark removed) as the `bytes` argument, whereas the binary path at `:575` passes the raw `bytes`.

Verified. For a `config.json` whose bytes are `EF BB BF 7B 7D`:

- `Kind="Binary"` produces `ConfigBytes => new byte[] { 0xEF, 0xBB, 0xBF, 0x7B, 0x7D }`
- `Kind="Both"` produces `ConfigBytes => new byte[] { 0x7B, 0x7D }`

Same file, same member name, different content, no diagnostic and no documentation of the difference.

The readme states flatly that "Binary files are embedded as-is" and scopes BOM removal to the Text section. Anyone who switches a file from `Binary` to `Both` to also get the text form silently changes the bytes underneath them — a hash comparison, a length assertion, or a byte-for-byte round trip to disk breaks with no visible cause. It also means `XBytes` is not a faithful copy of the file, which is the one property a "bytes" member should have.

## The change

The byte member now mirrors the file exactly for every `Kind`. Only the `const string` drops the BOM, which is what a C# consumer wants from text.

Also moves the binary size check from #2777 so it applies whenever the `Binary` flag is set, including `Both` — previously `Both` was bounded only by the text limit, and now that its byte member carries the raw bytes it should answer to the binary limit as well.

The readme's Text Encoding section is rewritten to state the rule directly rather than leaving it implied.

## Verification

New theory over `Binary` and `Both` asserting the byte member contains `0xEF, 0xBB, 0xBF, 0x7B, 0x7D` for both. The existing test that the text member drops the BOM is kept and renamed to say which member it is about.

23/23 pass.

### DIFF
--- a/src/Meziantou.Framework.EmbeddedConstantsGenerator/EmbeddedConstantsGeneratorTask.cs
+++ b/src/Meziantou.Framework.EmbeddedConstantsGenerator/EmbeddedConstantsGeneratorTask.cs
@@ -573,15 +573,15 @@ internal static class EmbeddedConstantsGeneratorTask
                 return new EmbeddedFile(file.FullPath, kind, file.ExplicitName, text: null, Array.Empty<byte>(), errors);
             }
 
+            // Every byte costs six characters of generated source, so a large file makes the build unusable
+            if (kind.HasFlag(EmbeddedConstantKind.Binary) && bytes.Length > maxBinaryFileBytes)
+            {
+                errors.Add(ValidationError.BinaryFileTooLarge(file.FullPath, maxBinaryFileBytes));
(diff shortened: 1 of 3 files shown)


meziantou:
Superseded by #2802, which is the same fix rebased onto `main` after #2776 and #2777 were closed. This PR still targets `fix/embeddedconstants-binary-size-limit`, the head of the closed #2777, so its diff no longer means anything. Closing.

PR #2779, opened 2026-08-30
Report member names that collide with each other or with the class name

meziantou:
**Stack 4 of 4** · merges into #2778 · must merge last

Four related defects in the member-naming pipeline. `CreateEntries` calls `UsePathBasedNamesForImplicitDuplicates` and then `ReportDuplicateMembers`, and those two methods sit one blank line apart (`:189` and `:191`) — they are one mechanism, so they are fixed together rather than as four PRs that all collide.

## Finding 1 — a member name equal to the class name emits uncompilable code, with no diagnostic

`:191-197`. `ReportDuplicateMembers` compares member names against each other, never against `options.ClassName`.

With `<EmbeddedConstantsClassName>FooText</EmbeddedConstantsClassName>` and `<EmbeddedConstant Include="Foo.txt" Kind="Text" />`, `Create` reports zero errors and emits:

```csharp
internal static partial class FooText
{
    public const string FooText = "x";
}
```

which fails with `error CS0542: 'FooText': member names cannot be the same as their enclosing type` (confirmed by compiling it). The user gets a Roslyn error pointing into a generated file they did not write, with no `MFECG` code to search for. Now reported as `MFECG0013`.

## Finding 2 — the name fallback embeds absolute filesystem paths

`:175-189` calling `MakeRelativePath:338-358`, which returns the full path unchanged when the file is not under `ProjectDirectory`.

Verified: two files named `shared.txt`, one in the project and one at `../shared/`, both without explicit `Name`. They collide, the fallback kicks in, and the out-of-project one becomes:

```csharp
public const string VarFoldersL_M8fk0x1s07ngvkcrk8mhhpqh0000gnTEcgOutsideSharedText = "o";
```

On CI the same file sits at a different absolute path, so the member is named something else entirely — code that references it compiles locally and fails on the build server. It also bakes the developer's home directory into shipped IL as a public member name.

Now the fallback is skipped for files outside the project directory, so the collision survives and is reported as `MFECG0005` telling the user to set `Name`. An error is better than a name that depends on where the repository is checked out.

## Finding 3 — `MFECG0005` named the member but not the files

`:658-661`, with `FilePath` passed as `null` at `:195`. With a glob over a large asset tree, finding the two culprits was manual work. The message now lists the participating files and the error is attributed to the first of them so the IDE can navigate to it.

## Finding 4 — `MakeRelativePath` compared paths case-insensitively

`:347` used `StringComparison.OrdinalIgnoreCase`. On Linux, `/src/Project/` and `/src/project/` are different directories, but this treated a file under the latter as being under the former. Replaced with `Path.GetRelativePath`, which applies the current platform's rules and also normalises separators; a result that is rooted or starts with a `..` segment means "not under the base path".

## Verification

Three new unit tests: the class-name collision reports `MFECG0013`; the out-of-project collision reports `MFECG0005` rather than inventing a path-based name; the duplicate message names both files and sets `FilePath`. The existing test that in-project duplicates still get path-based names (`ASettingsText` / `BSettingsText`) continues to pass, which is what keeps finding 2's fix from over-reaching.

26/26 pass on `net10.0` and `net11.0`.

### DIFF
--- a/src/Meziantou.Framework.EmbeddedConstantsGenerator/EmbeddedConstantsGeneratorTask.cs
+++ b/src/Meziantou.Framework.EmbeddedConstantsGenerator/EmbeddedConstantsGeneratorTask.cs
@@ -168,7 +168,7 @@ internal static class EmbeddedConstantsGeneratorTask
         }
 
         UsePathBasedNamesForImplicitDuplicates(options, entries);
-        ReportDuplicateMembers(entries, errors);
+        ReportInvalidMemberNames(options, entries, errors);
 
         return entries;
     }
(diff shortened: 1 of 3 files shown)


meziantou:
Superseded by #2803, which is the same fix rebased onto `main` after #2776 and #2777 were closed. This PR still targets `fix/embeddedconstants-bom-consistency`, so its diff no longer means anything. Closing.

PR #2804, opened 2026-08-31
Bump the shipping package version when only its analyzer changes

meziantou:
## Problem

`eng/update-versions.cs` skips every non-packable project before it builds the dependency graph:

```csharp
var isPackable = IsPackableProject(file);
if (!isPackable)
{
    continue;   // never reaches allCsprojFiles / changesPerCsproj / the graph
}
```

`Meziantou.Framework.Assertions.Analyzer` and `.CodeFix` both set `IsPackable=false` — they ship *inside* the `Meziantou.Framework.Assertions` package via `<None Include ... PackagePath="analyzers/dotnet/cs" />`, and the package declares that dependency explicitly with `ProjectReference` + `ReferenceOutputAssembly="false"`.

`GetProjectReferences` already reads every `ProjectReference` regardless of `ReferenceOutputAssembly`, so the dependency was always visible to the script. But because the analyzer was never added to the graph, `reverseDependencyGraph[analyzer]` did not exist, so a commit touching only `src/Meziantou.Framework.Assertions.Analyzer/**`:

1. mapped via `GetCsproj()` to the analyzer csproj,
2. failed the `packableProjects.Contains(csproj)` check and was dropped,
3. and was never picked up by the transitive-dependents pass either.

The version stayed put, `create_nuget` repacked a package containing the **new** analyzer DLL, and the push ran with `--skip-duplicate`, which silently no-ops against the already-published version. CI stayed green and the fix never reached users.

## Change

Follow the explicit `ProjectReference`, without touching the existing graph or passes.

- One new map, built from the packable projects' own `ProjectReference`s: non-packable project → the packable projects that ship it.
- One resolver, `GetVersionedProjects(csproj)`: the project itself when it is packable, otherwise the packages that reference it. Recursive, so an analyzer sitting behind another non-packable project still resolves.
- The two change-attribution loops resolve through it.

`allCsprojFiles`, the dependency graph, the direct-change pass and the transitive-dependent pass are all unchanged. A non-packable project can never enter `projectsToUpdate`, so it never gains a `<Version>` element. Test projects keep the previous behavior: `GetVersionedProjects` returns an empty list for a project that is neither packable nor shipped by one.

The analyzer's commits are attributed **directly** to the shipping package, so the generated bump PR lists the real commit messages rather than `Updated due to dependency on ...`.

## Scope

I checked every package that packs an analyzer. FullPath, FastEnumGenerator and Yaml all `ProjectReference` what they pack; ResxSourceGenerator, FixedStringBuilder.Generator and ServiceDefaults.AutoRegister pack their own `$(AssemblyName).dll` and so have no cross-project dependency to declare. No csproj needs a new `ProjectReference`.

## Verification

Synthetic analyzer-only commit, run with `--since "10 minutes ago"` and no `--create-pull-request`:

| | `Meziantou.Framework.Assertions` version |
| --- | --- |
| `main` | `2.0.5` — unchanged |
| this branch | **`2.0.6`**, listed with the analyzer's own commit |

Both `Meziantou.Framework.Assertions.Analyzer.csproj` and `.CodeFix.csproj` still contain zero `<Version>` elements, confirming non-packable projects are not versioned. `dotnet build eng/update-versions.cs` succeeds.

Found by a project review of `Meziantou.Framework.Assertions`.

### DIFF
--- a/eng/update-versions.cs
+++ b/eng/update-versions.cs
@@ -56,15 +56,8 @@ var changesPerCsproj = new Dictionary<string, CsprojInfo>(StringComparer.Ordinal
 var allCsprojFiles = new List<string>();
 var packableProjects = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
-// Every project under src, packable or not. Non-packable projects (analyzers, code fixes, source
-// generators) ship inside a packable package, so they must take part in the dependency graph even
-// though they never carry a version of their own.
-var allProjectFiles = new List<string>();
-
 foreach (var file in Directory.EnumerateFiles(srcPath, "*.csproj", SearchOption.AllDirectories))
 {
-    allProjectFiles.Add(file);
-
     var isPackable = IsPackableProject(file);
     if (isPackable)
     {
@@ -86,13 +79,13 @@ Console.WriteLine("Building dependency graph...");
 var dependencyGraph = new Dictionary<string, List<string>>(StringComparer.OrdinalIgnoreCase);
 var reverseDependencyGraph = new Dictionary<string, List<string>>(StringComparer.OrdinalIgnoreCase);
 
-foreach (var csproj in allProjectFiles)
+foreach (var csproj in allCsprojFiles)
 {
     dependencyGraph[csproj] = GetProjectReferences(csproj);
     reverseDependencyGraph[csproj] = [];
 }
 
-foreach (var csproj in allProjectFiles)
+foreach (var csproj in allCsprojFiles)
 {
     foreach (var dependency in dependencyGraph[csproj])
     {
@@ -100,6 +93,27 @@ foreach (var csproj in allProjectFiles)
         {
             list.Add(csproj);
         }
+    }
+}
+
+// A project that is not packable (an analyzer, a code fix, a source generator) carries no version of its own: it
+// ships inside the packable projects that reference it with ProjectReference/ReferenceOutputAssembly=false. Map it
+// to those packages so that a change confined to it still bumps the version of the package that ships it.
+var shippedByPackages = new Dictionary<string, List<string>>(StringComparer.OrdinalIgnoreCase);
+foreach (var csproj in allCsprojFiles)
+{
+    foreach (var dependency in GetProjectReferences(csproj))
+    {
+        if (packableProjects.Contains(dependency))
+            continue;
+
+        if (!shippedByPackages.TryGetValue(dependency, out var packages))
+        {
+            packages = [];
+            shippedByPackages[dependency] = packages;
+        }
+
+        packages.Add(csproj);
     }
 }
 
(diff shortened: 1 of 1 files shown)


meziantou:
I don't think this is the right fix. A dependency between 2 projects must be explicit (ProjectReference, with ReferenceAssembly=false if needed). So, the bump-version logic should handle this case.

meziantou:
You're right that the dependency has to be explicit, and it already is — `Meziantou.Framework.Assertions.csproj` has `ProjectReference` to both the analyzer and the code fix with `ReferenceOutputAssembly="false"`. I checked the other packages that pack an analyzer and they're fine too: FullPath, FastEnumGenerator and Yaml all reference what they pack, and ResxSourceGenerator, FixedStringBuilder.Generator and ServiceDefaults.AutoRegister pack their *own* `$(AssemblyName).dll`, so there is no cross-project dependency to declare. Nothing needs a new `ProjectReference`.

`GetProjectReferences` already reads every `ProjectReference` regardless of `ReferenceOutputAssembly`, so the explicit dependency was always visible to the script. The only reason it wasn't followed is that non-packable projects were skipped before the graph was built, so `reverseDependencyGraph[analyzer]` never existed.

My first commit did follow that `ProjectReference`, but it got there by widening the graph to every project and then re-filtering in three places, which doesn't read like "follow the explicit reference" at all. Fair call. I've pushed a second commit that reworks it:

- `allCsprojFiles`, the dependency graph, and both the direct-change and transitive-dependent passes are now **untouched**.
- One new map, built straight from the packable projects' `ProjectReference`s: non-packable project → the packable projects that ship it.
- One resolver, `GetVersionedProjects(csproj)`: the project itself when packable, otherwise the packages that reference it (recursive, so an analyzer behind another non-packable project still resolves).
- The two attribution loops resolve through it.

Net effect is +62/−16 concentrated in one concept instead of a restructuring.

One behavior change worth noting, which I think is an improvement: the analyzer's commits are now attributed **directly** to the shipping package, so the generated bump PR lists the real commit messages instead of `Updated due to dependency on Meziantou.Framework.Assertions.Analyzer`.

Verified the same way as before — a synthetic analyzer-only commit, run with `--since "10 minutes ago"` and no `--create-pull-request`:

| | `Meziantou.Framework.Assertions` |
| --- | --- |
| `main` | `2.0.5` — unchanged |
| this branch | **`2.0.6`**, listed with the analyzer's own commit |

`Meziantou.Framework.Assertions.Analyzer.csproj` and `.CodeFix.csproj` still contain zero `<Version>` elements, so non-packable projects are never versioned. Test projects are unaffected: `GetVersionedProjects` returns an empty list for a project that is neither packable nor shipped by one, which preserves the old skip behavior.

If you meant something different — e.g. that the version should be driven off the pack items rather than `ProjectReference`, or that you'd rather this stayed in the transitive-dependent pass — say so and I'll redo it.

PR #2857, opened 2026-08-31
Implement ISupportExternalScope on XUnitLoggerProvider

meziantou:
Scopes the logger factory synthesises from the current `Activity` never reached the xUnit output.

### The problem

`XUnitLoggerProvider` implemented `ILoggerProvider` only, and each instance owned a private `LoggerExternalScopeProvider` (`XUnitLoggerProvider.cs:27`). `LoggerFactory` injects `TraceId`/`SpanId` scopes into its *own* scope provider and hands that out only to providers implementing `ISupportExternalScope`, so this provider never received it.

Measured with `ActivityTrackingOptions.SpanId | ActivityTrackingOptions.TraceId`, `AddXunit(output, new XUnitLoggerOptions { IncludeScopes = true })` and a started `Activity`, the emitted line was exactly:

```
act
```

No `TraceId`, no `SpanId`.

Scopes created explicitly through `ILogger.BeginScope` were unaffected and did appear — the composite `Logger` pushes those into each provider's logger directly — which is what made this easy to miss.

This matters in the scenario the package exists for: running an ASP.NET Core host under `WebApplicationFactory` and reading its logs in a test. That is exactly where correlation ids matter, and a `Console` provider running alongside *will* show them, so the discrepancy looks unexplainable.

### What changed

Mirrors #2642 (`Implement ISupportExternalScope on InMemoryLoggerProvider`), including the `MutableExternalScopeProvider` indirection: the factory hands over its scope provider only after the provider is registered, so loggers hold the indirection rather than a fixed `IExternalScopeProvider` and a logger created before the hand-off still observes the factory's scopes afterwards.

`XUnitLogger._scopeProvider` widens from `LoggerExternalScopeProvider` to `IExternalScopeProvider`, with a new `internal` constructor taking the interface. **All three public constructors keep their existing `LoggerExternalScopeProvider` parameter**, so there is no public API break — the `ref/` diff is limited to the added interface and `SetScopeProvider`.

### Verification

- `ActivityTrackingScopesAreWrittenToTheOutput` asserts the record contains the activity's `TraceId` and `SpanId`. It fails on `main` and passes here.
- `ScopesStartedByTheCallerAreStillWrittenToTheOutput` is a regression guard for ordinary `BeginScope` output; it passes both before and after, confirming the widening does not disturb the existing path.

`dotnet test` — 10/10 across `net10.0` and `net11.0`. `ref/` regenerated with a Release `IsOfficialBuild` build; `dotnet run ./eng/update-all.cs` produces no further changes.

---

**Stack 2 of 2** · merges into `fix/xunit-provider-alias` · #2856 must merge first.

### DIFF
--- a/ref/Meziantou.Extensions.Logging.Xunit.v3.cs
+++ b/ref/Meziantou.Extensions.Logging.Xunit.v3.cs
@@ -38,13 +38,14 @@ namespace Meziantou.Extensions.Logging.Xunit.v3
         public bool UseUtcTimestamp { get => throw null; set { } }
     }
 
-    public sealed class XUnitLoggerProvider : Microsoft.Extensions.Logging.ILoggerProvider, System.IDisposable
+    public sealed class XUnitLoggerProvider : Microsoft.Extensions.Logging.ILoggerProvider, Microsoft.Extensions.Logging.ISupportExternalScope, System.IDisposable
     {
         public XUnitLoggerProvider(Xunit.ITestOutputHelper? testOutputHelper) { }
         public XUnitLoggerProvider(Xunit.ITestOutputHelper? testOutputHelper, bool appendScope) { }
         public XUnitLoggerProvider(Meziantou.Extensions.Logging.Xunit.v3.XUnitLoggerOptions? options) { }
         public XUnitLoggerProvider(Xunit.ITestOutputHelper? testOutputHelper, Meziantou.Extensions.Logging.Xunit.v3.XUnitLoggerOptions? options) { }
         public Microsoft.Extensions.Logging.ILogger CreateLogger(string categoryName) => throw null;
+        public void SetScopeProvider(Microsoft.Extensions.Logging.IExternalScopeProvider scopeProvider) { }
         public void Dispose() { }
     }
 }
--- a/src/Meziantou.Extensions.Logging.Xunit.v3/MutableExternalScopeProvider.cs
+++ b/src/Meziantou.Extensions.Logging.Xunit.v3/MutableExternalScopeProvider.cs
@@ -1,0 +1,29 @@
+using Microsoft.Extensions.Logging;
+
+#pragma warning disable IDE1006 // Naming Styles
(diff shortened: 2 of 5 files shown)


meziantou:
Closing — `main` already has this, and its implementation is better than mine.

I reviewed both rather than assuming. `XUnitLoggerProvider` on `main` already implements `ISupportExternalScope` with `Volatile` read/write on both the provider field and `XUnitLogger.ScopeProvider`, and it does two things this PR does not:

- **It caches loggers** in a `ConcurrentDictionary<string, XUnitLogger>`, so a category gets a stable logger instance instead of a fresh allocation per `CreateLogger` call. Merging this PR would have removed that.
- **It exposes a public `XUnitLogger(ITestOutputHelper?, IExternalScopeProvider, string?, XUnitLoggerOptions?)` constructor**, where I had kept the equivalent `internal`. The public one is the more useful call.

It is also already covered by `SetScopeProviderUpdatesExistingAndNewLoggers` and `LoggerFactorySetsTheScopeProvider`, so my two tests add nothing — and the first of those covers propagation to *already-created* loggers, which my tests did not.

The one difference in my approach was a `MutableExternalScopeProvider` indirection (mirroring #2642 on `InMemoryLoggerProvider`) instead of `main`'s propagation loop. That indirection is immune to a create/set interleaving that `main`'s loop is not: a logger constructed with the old provider on one thread, inserted into `_loggers` after `SetScopeProvider`'s `foreach` has already passed, would keep the stale provider.

I checked whether that is actually reachable rather than reporting it as a defect, and it is not through the normal path: `LoggerFactory.CreateLogger` and `LoggerFactory.AddProvider` (which calls `SetScopeProvider` from `AddProviderRegistration`) take the same `_sync` lock, so they are serialised. It would need someone driving `XUnitLoggerProvider.CreateLogger` and `SetScopeProvider` concurrently by hand. Not worth a change; noting it only so the trade-off is on the record.

#2856 (the `ProviderAlias`, which `main` still does not have) stays open and now targets `main` directly.

Issue #2952, opened 2026-09-03
`Meziantou.Framework.Scheduling` with `DateTimeOffset`


viceroypenguin:
I am starting to integrate `Meziantou.Framework.Scheduling`, but I notice that `.GetNextOccurrence()` (and it's supporting method `.GetNextOccurrences()`) only support `DateTime`.

Would it be possible to ask you to support `.GetNextOccurrence(DateTimeOffset, TimeZoneInfo)` or `.GetNextOccurrence(DateTimeOffset, string timeZone)`?  In this case, it would need to properly support ambiguous times, such as `@minutely` through a DST overlap back window.

meziantou:
Adding DateTimeOffset should be pretty easy. However, adding `TimeZoneInfo` means answering many questions about ambiguous or invalid dates. I don't know if there is a common solution for it.
What should be the result of GetNextOccurence with a timezone when the next occurrence is not a valid date, or when the date is ambiguous? Can you explain your use case?

viceroypenguin:
> However, adding `TimeZoneInfo` means answering many questions about ambiguous or invalid dates. I don't know if there is a common solution for it.

Agreed; I don't know if there is a common global standard. Given that this is for a task scheduler, it is important to me, at least, to figure out if there's an answer. I'll do some research and see if I can present a general solution, or if this is something I should handle myself.

viceroypenguin:
If I give you an algorithm that returns every DateTimeOffset occurrence during a dst window, would that be acceptable? Then consumers can decide if they want to exclude them based on rules, but at least the possibilities are available to work with. Doing so would also deal with changing offsets through the window, in either direction. As far as missed occurrences due to jumping forward, I think maybe we just give them the start of the DST period and jump forward from there?

meziantou:
I haven't looked into the details yet, but it seems RFC 5545 (RRule) define timezones and how to handle them. Cron expressions don't support timezone. They are some unofficial extensions for it.

Side note: Time zones are often defined using iana format (e.g. `america/toronto`) which is only supported on .NET 8+ (actually a previous .NET Core, but they are not supported anymore). So, the netstandard2.0 version may not be able to use them as-is.

viceroypenguin:
> Side note: Time zones are often defined using iana format (e.g. america/toronto) which is only supported on .NET 8+ (actually a previous .NET Core, but they are not supported anymore). So, the netstandard2.0 version may not be able to use them as-is.

[...]

PR #2953, opened 2026-09-03
Add a DateTimeOffset overload of GetNextOccurrence

meziantou:
Contributes to https://github.com/meziantou/Meziantou.Framework/issues/2952

## What

Adds a `DateTimeOffset` overload to `RecurrenceRuleExtensions.GetNextOccurrence`:

```csharp
public static DateTimeOffset? GetNextOccurrence(this IRecurrenceRule recurrenceRule, DateTimeOffset startDate)
```

It applies to every `IRecurrenceRule`, so both `RecurrenceRule` (RFC 5545) and `CronExpression` get it.

## Why

Callers holding a `DateTimeOffset` previously had to unwrap it, call the `DateTime` overload, and rewrap the result themselves — deciding the offset semantics along the way, which is exactly the part that is easy to get wrong.

## Semantics

Occurrences are computed from the **wall-clock time** of `startDate` (`startDate.DateTime`), and the result carries the **same offset** as `startDate`.

This keeps RFC 5545 fields bound to local time, as the spec intends: `BYHOUR=10` means 10:00 in the caller's offset, not 10:00 UTC. Cron expressions are conventionally local too. A caller who wants UTC evaluation passes a zero offset (`startDate.ToOffset(TimeSpan.Zero)`), so the offset itself is the knob.

Computing on `startDate.UtcDateTime` instead and reapplying the original offset would shift the instant by that offset — the first occurrence of a daily rule would no longer equal the start date.

`DateTime.SpecifyKind(..., Unspecified)` guards the `DateTimeOffset(DateTime, TimeSpan)` constructor, which throws when a `Utc`-kind value is paired with a non-zero offset, in case a rule implementation returns a kinded value.

## Notes for reviewers

- Three tests in `RecurrenceRuleTests`. They assert `.Offset` explicitly rather than relying on `Assert.Equal`, since `DateTimeOffset.Equals` compares instants only and would pass even if the offset were dropped.
- `ref/Meziantou.Framework.Scheduling.cs` regenerated; the new API is present on all three TFMs, netstandard2.0 included.
- Package version bumped 4.0.4 → 4.0.5, matching the repo's patch-bump convention (the netstandard2.0 target addition was also a patch bump). Say the word if this additive API warrants 4.1.0 instead.

Verified: Release build with `TreatsWarningsAsErrors=true` clean, `dotnet run ./eng/update-all.cs` produces no further changes, and the full Scheduling suite passes 630/630 on net10.0 and net11.0.

### DIFF
--- a/src/Meziantou.Framework.Scheduling/RecurrenceRuleExtensions.cs
+++ b/src/Meziantou.Framework.Scheduling/RecurrenceRuleExtensions.cs
@@ -14,4 +14,18 @@ public static class RecurrenceRuleExtensions
 
         return null;
     }
+
+    /// <summary>Gets the next occurrence of the recurrence starting from the specified date.</summary>
+    /// <param name="startDate">The date to start searching for the next occurrence.</param>
+    /// <returns>The next occurrence date with the same offset as <paramref name="startDate"/>, or <see langword="null"/> if there are no more occurrences.</returns>
+    /// <remarks>The occurrences are computed using the local time of <paramref name="startDate"/>, and the offset of <paramref name="startDate"/> is applied to the result.</remarks>
+    public static DateTimeOffset? GetNextOccurrence(this IRecurrenceRule recurrenceRule, DateTimeOffset startDate)
+    {
+        ArgumentNullException.ThrowIfNull(recurrenceRule);
+
+        foreach (var occurrence in recurrenceRule.GetNextOccurrences(startDate.DateTime))
+            return new DateTimeOffset(DateTime.SpecifyKind(occurrence, DateTimeKind.Unspecified), startDate.Offset);
+
+        return null;
+    }
 }
--- a/ref/Meziantou.Framework.Scheduling.cs
+++ b/ref/Meziantou.Framework.Scheduling.cs
@@ -101,6 +101,7 @@ namespace Meziantou.Framework.Scheduling
     public static class RecurrenceRuleExtensions
     {
         public static System.DateTime? GetNextOccurrence(this Meziantou.Framework.Scheduling.IRecurrenceRule recurrenceRule, System.DateTime startDate) => throw null;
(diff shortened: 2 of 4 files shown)


src/Meziantou.Framework.Scheduling/RecurrenceRuleExtensions.cs
```diff
@@ -14,4 +14,18 @@ public static class RecurrenceRuleExtensions
...
+    /// <summary>Gets the next occurrence of the recurrence starting from the specified date.</summary>
+    /// <param name="startDate">The date to start searching for the next occurrence.</param>
+    /// <returns>The next occurrence date with the same offset as <paramref name="startDate"/>, or <see langword="null"/> if there are no more occurrences.</returns>
+    /// <remarks>The occurrences are computed using the local time of <paramref name="startDate"/>, and the offset of <paramref name="startDate"/> is applied to the result.</remarks>
+    public static DateTimeOffset? GetNextOccurrence(this IRecurrenceRule recurrenceRule, DateTimeOffset startDate)
+    {
+        ArgumentNullException.ThrowIfNull(recurrenceRule);
+
+        foreach (var occurrence in recurrenceRule.GetNextOccurrences(startDate.DateTime))
+            return new DateTimeOffset(DateTime.SpecifyKind(occurrence, DateTimeKind.Unspecified), startDate.Offset);
```
viceroypenguin, line 27: Per the conversation in #2952, this will definitely be an incorrect answer, in the context of offset changes during dst changes.

meziantou: DateTimeOffset doesn't carry the time zone information, only the offset, so I think this implementation works fine in this context.
If there is a good way to support timezone in this library, it will definitly be another overload

viceroypenguin: Fair

PR #2981, opened 2026-09-05
Support volumes in TemporaryContainers

meziantou:
Volumes in `Meziantou.Framework.TemporaryContainers` were declarative only: a `VolumeMount` name was passed through to the runtime and nothing else happened. The library never created or deleted a volume, so a test had to manage the volume's lifetime itself, and `docker rm -f` ran without `-v`, leaking the image's anonymous volumes on every run.

## Volume lifecycle

- **`VolumeDefinition`** (`Runtime`, `Name`, `Driver`, `ReuseId`, `Labels`, `DriverOptions`, deep-copy ctor, `CreateVolume()`) and **`TemporaryVolume : IAsyncDisposable`** (`Name`, `EnsureCreatedAsync`, `ExistsAsync`, `DeleteAsync`), mirroring `ContainerDefinition` / `TemporaryContainer`. The name is resolved eagerly because Apple's `container volume create` takes it as a mandatory positional argument.
- `EnsureCreatedAsync` probes before creating, so it records ownership **only when it actually created the volume**. Pointing a definition at an existing volume adopts it and dispose leaves it alone — without that probe, `VolumeDefinition.Name = "my-precious-data"` would be destroyed on dispose, since `docker volume create <existing>` is idempotent. The same probe makes creation idempotent on podman and Apple, which reject an existing name.
- `DeleteAsync` re-probes and throws when the volume survived: every runtime reports "still in use" the same way it reports "already gone", so the outcome is confirmed rather than inferred from an exit code.
- `Mounts.AddVolume(TemporaryVolume, path, readOnly)` records the volume and the container creates it on start. A volume bound to a different engine is rejected — otherwise the mismatch surfaces as a mysteriously empty directory rather than an error.
- New volume verbs on `ContainerRuntime`, implemented for the docker/podman CLI, the Docker Engine API and Apple's CLI, and forwarded by `Auto`.

## Fixes along the way

- **`VolumeMount` gains `ReadOnly`**, like `BindMount`. This rewrites the record's primary constructor and `Deconstruct`, so it is binary-breaking and the package version goes to **1.1.0**. The existing 2-arg `AddVolume` is left untouched and a 3-arg overload added, to avoid a second break.
- Anonymous volumes are removed with the container (`rm -f -v`, `?force=1&v=1`). Named volumes are never touched by either flag.
- `--mount` fields containing `,` or `"` are CSV-quoted for docker, and rejected with a clear error on the runtimes whose parsers split on the comma, instead of silently emitting a corrupt descriptor. Both behaviours were checked against the real CLIs.
- Apple's `container` 1.1.0 supports `--tmpfs`, `--read-only`, `--network` and `--mount type=volume,…,readonly`, so those stale `NotSupportedException`s are gone and its mounts now use `--mount` like the other runtimes. Only `--network-alias` still throws — it genuinely does not exist.
- `docker volume rm` is used without `--force`: on docker it only hides the not-found error (already tolerated), while `podman volume rm --force` additionally removes the containers using the volume.

## Reviewer notes

- **`wslc` has no volume commands** and throws. `AutoContainerRuntime` yields `Wslc` first on Windows and caches the resolution, so `ContainerRuntime.Auto` will throw for volume operations on a Windows box with `wslc` even when Docker Desktop is installed. Noted in the readme.
- **Apple runtime limitation:** `container` 1.1.0 hangs indefinitely on any operation (`exec`, `cp`, even `delete -f`) against a container that mounts a volume a *deleted* container previously used, and the daemon degrades badly. `AppleContainerContainerTests` therefore runs the single-container volume lifecycle test but not the cross-container one. Documented in the readme.
- `OwnedVolumeMount` is internal and correctly absent from `ref/`; the rest of the `ref/` diff is the intended new public surface.

## Verification

- `dotnet build` in Release with `IsOfficialBuild=true` and `TreatWarningsAsErrors=true`; `ref/` regenerated and committed. `dotnet run ./eng/update-all.cs` produced no further changes. Dependent projects (`Trimmable`, `Http.Caching.Tests`) build.
- Full test suite on net10.0 and net11.0: **268 passed, 48 skipped** (podman and wslc are not installed on the dev machine), **4 failed** — the two `SqlServerContainerTests` × 2 TFMs, failing on `image … does not support required platforms` because there is no arm64 mssql image for Apple's runtime. Pre-existing on that machine and unrelated to this change.
- All volume tests pass on both TFMs, including the docker and Apple integration tests and the Docker Engine API volume endpoints. Docker volume count was identical before and after the run, and no Apple volumes were left behind.

### DIFF
--- a/ref/Meziantou.Framework.TemporaryContainers.cs
+++ b/ref/Meziantou.Framework.TemporaryContainers.cs
@@ -134,6 +134,8 @@ namespace Meziantou.Framework.TemporaryContainers
         public void Add(Meziantou.Framework.TemporaryContainers.IMount mount) { }
         public void AddBindMount(string hostPath, string containerPath, bool readOnly = false) { }
         public void AddVolume(string volumeName, string containerPath) { }
+        public void AddVolume(string volumeName, string containerPath, bool readOnly) { }
+        public void AddVolume(Meziantou.Framework.TemporaryContainers.TemporaryVolume volume, string containerPath, bool readOnly = false) { }
         public void AddTmpfs(string containerPath) { }
         public System.Collections.Generic.IEnumerator<Meziantou.Framework.TemporaryContainers.IMount> GetEnumerator() => throw null;
         System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator() => throw null;
@@ -371,6 +373,17 @@ namespace Meziantou.Framework.TemporaryContainers
         public System.Collections.Generic.IAsyncEnumerable<Meziantou.Framework.TemporaryContainers.LogEntry> GetLogsAsync([System.Runtime.CompilerServices.EnumeratorCancellation] System.Threading.CancellationToken cancellationToken = null) => throw null;
     }
 
+    public class TemporaryVolume : System.IAsyncDisposable
+    {
+        public string Name { get => throw null; }
+        public Meziantou.Framework.TemporaryContainers.VolumeDefinition Definition { get => throw null; }
+        public Meziantou.Framework.TemporaryContainers.ContainerRuntime Runtime { get => throw null; }
+        public System.Threading.Tasks.Task EnsureCreatedAsync(System.Threading.CancellationToken cancellationToken = null) => throw null;
+        public System.Threading.Tasks.Task<bool> ExistsAsync(System.Threading.CancellationToken cancellationToken = null) => throw null;
+        public System.Threading.Tasks.Task DeleteAsync(System.Threading.CancellationToken cancellationToken = null) => throw null;
+        public System.Threading.Tasks.ValueTask DisposeAsync() => throw null;
+    }
+
     public sealed class TmpfsMount : Meziantou.Framework.TemporaryContainers.IMount, System.IEquatable<Meziantou.Framework.TemporaryContainers.TmpfsMount>
     {
         public string Target { get => throw null; init { } }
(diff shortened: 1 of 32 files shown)


ref/Meziantou.Framework.TemporaryContainers.cs
```diff
@@ -372,6 +374,17 @@ public class TemporaryContainer : System.IAsyncDisposable
         public System.Collections.Generic.IAsyncEnumerable<Meziantou.Framework.TemporaryContainers.LogEntry> GetLogsAsync([System.Runtime.CompilerServices.EnumeratorCancellation] System.Threading.CancellationToken cancellationToken = null) => throw null;
     }
 
+    public class TemporaryVolume : System.IAsyncDisposable
```
meziantou: Should it be sealed?

ref/Meziantou.Framework.TemporaryContainers.cs
```diff
@@ -385,18 +398,41 @@ public TmpfsMount(string Target) { }
         public void Deconstruct(out string Target) => throw null;
     }
 
+    public class VolumeDefinition
```
meziantou: Should it be sealed?

ref/Meziantou.Framework.TemporaryContainers.cs
meziantou: Yes — sealed in 383f7c26a.

`TemporaryContainer` is open only because the database helpers derive from it (`RedisContainer`, `PostgreSqlContainer`, `SqlServerContainer`, `MongoDbContainer`). There is no volume equivalent and nothing derives from `TemporaryVolume`, so leaving it open was shape-matching rather than a requirement. Unsealing later is not a breaking change; sealing later is.

ref/Meziantou.Framework.TemporaryContainers.cs
meziantou: Yes — sealed in 383f7c26a.

Same reasoning as `TemporaryVolume`: `ContainerDefinition` is open for `RedisContainerDefinition` and friends, and volumes have no such helpers.

One knock-on: a sealed type cannot declare a virtual member, so `CreateVolume()` lost its `virtual`. It only had it to mirror `ContainerDefinition.CreateContainer()`, which needs it for those subclasses.

PR #2983, opened 2026-09-05
Run the wslc container tests on the Windows CI agents

meziantou:
## What

- `ContainerRuntimeTestsBase` gains an opt-in `IsRuntimeRequired` gate. When a runtime is declared mandatory, an unavailable runtime **fails** the tests instead of skipping them.
- `WslcContainerTests` opts in on GitHub Actions Windows only (`OperatingSystem.IsWindows() && TestEnvironment.IsOnGitHubActions()`).
- CI installs WSL from the pre-release channel (`wsl --update --pre-release`) on the Windows leg of the `TemporaryContainers` test group, so the tests have a runtime to drive.
- New `debug_wslc` job reports `wsl --version` and `wslc version` on `windows-2025-vs2026`, independently of the test run.

## Why

The Wslc tests were collected on all four matrix legs but always skipped: the suite gate skips when the runtime does not answer, and no hosted agent ships `wslc`. A runtime that silently never runs looks exactly like one that passes, so the suite gave no signal on Windows. Every other runtime and every local run keep the previous skip behavior.

## Notes for reviewers

- **This may go red on the first run, by design.** GitHub's hosted Windows runners have historically lacked nested virtualization, which WSL2 (and therefore `wslc`) needs. If the pre-release install does not produce a working `wslc`, the 10 Wslc tests fail on the Windows leg and `debug_wslc` fails — that is the signal this PR is meant to produce, and the debug job's output distinguishes "wslc not installed" from "WSL2 cannot start here".
- The install step's `wsl --version` / `wslc version` calls are deliberately non-fatal (try/catch + `exit 0`) so a missing `wslc` surfaces as test failures rather than a broken setup step. The `debug_wslc` job's `wslc version` step, in contrast, does fail the job.
- `debug_wslc` was added to `final_status_check`'s `needs` so it cannot fail unnoticed. Happy to drop that if you'd rather keep it purely informational.

## Verification

- Build clean with `/p:TreatsWarningsAsErrors=true`, 0 warnings.
- Forced the required path locally (temporarily dropping the Windows condition, `GITHUB_ACTIONS=true`): 20 failures instead of 20 skips, confirming the tests do fail when `wslc` is missing. Reverted afterwards.
- Full `Meziantou.Framework.TemporaryContainers.Tests` run on macOS/arm64: 266 total, 220 passed, 42 skipped, 4 failed — the known SqlServer-on-arm64 baseline, unchanged by this PR.
- `dotnet run ./eng/update-all.cs` — exit 0, no generated-file changes.
- The CI behavior itself could not be verified locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -419,18 +419,10 @@ jobs:
                 Enable-WindowsOptionalFeature -Online -FeatureName Client-ProjFS -NoRestart
               }
 
-          # wslc ships with WSL, and only the pre-release channel provides it. The Wslc tests fail instead of
-          # skipping on Windows agents, so this step is what keeps them running.
-          - name: Install WSL containers (wslc)
+          # The Wslc tests fail instead of skipping on Windows agents, so this step is what keeps them running
+          - name: Setup wslc
             if: startsWith(matrix.runs-on, 'windows') && matrix.test-group.id == 'TemporaryContainers' && steps.build-project.outputs.has-project == 'true'
-            run: |
-              wsl --update --pre-release
-              if ($LASTEXITCODE -ne 0) { throw "wsl --update --pre-release failed with exit code $LASTEXITCODE" }
-
-              # Diagnostics only: a missing wslc must surface as a Wslc test failure, not as a broken setup step
-              try { wsl --version } catch { Write-Host "::warning::wsl --version failed: $_" }
-              try { wslc version } catch { Write-Host "::warning::wslc version failed: $_" }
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -454,6 +459,17 @@ jobs:
               path: |
                 **/*.coverage
 
+  # Debug job: reports which wslc the Windows agents provide, independently of the test run
+  debug_wslc:
```
meziantou, line 463: To remove

PR #2985, opened 2026-09-05
Require the docker, Docker API and podman container tests on the CI agents

meziantou:
## What changed

**Removed the `debug_wslc` job.** It reported which `wslc` the Windows agents provide while that runtime was being brought up. The `Wslc` tests cover it now, so the job is gone from `ci.yml` along with its entry in `final_status_check`'s `needs` list, and the comment in `setup-wslc/action.yml` that pointed at it is updated.

**Gated docker, Docker API and podman the way wslc already is.** `WslcContainerTests` overrides `IsRuntimeRequired` so an unavailable runtime *fails* instead of skipping — a broken CI setup cannot silently turn the suite into a no-op. The docker and podman suites had no such gate, and the Docker Engine API runtime had no integration test at all, only log-parsing unit tests.

- `DockerContainerTests` and `PodmanContainerTests` now require their runtime on the Linux CI agents.
- New `DockerApiContainerTests` runs the shared suite against the Docker Engine API runtime, required on the Linux CI agents.

The `ubuntu-26.04` runner images ship Docker 29.4.2 (client + server) and Podman 5.7.0 on **both** x64 and arm64, so unlike wslc no setup step is needed — only the gate was missing.

The Windows agents are deliberately left out. Their docker fails intermittently for reasons unrelated to the change under test (the retry workflow usually turns those runs green on the second attempt), so requiring it there would only add red.

## Docker Engine API runtime fixes

Running the shared suite against that runtime for the first time exposed three real defects:

| | |
|---|---|
| **`start`/`stop` failed on a container already in the requested state** | The daemon reports that with `304 Not Modified`, which was treated as an error, so adopting a reused container threw. `SendAsync`'s `allowNotFound` flag becomes a general `allowedStatusCode`. |
| **Dockerfile builds threw `NotSupportedException`** | Now `POST /build` with a tar of the build context, including the case of a Dockerfile stored outside of it. |
| **All four file operations threw `NotSupportedException`** | Now `GET`/`PUT /containers/{id}/archive`, with a `HEAD` to stat the target so the copy-into-a-directory vs. rename semantics of `docker cp` are matched. |

New `DockerApiTarArchive` does the tar work with `System.Formats.Tar`, and rejects archive entries that escape the destination directory.

This matters beyond CI: `ContainerRuntime.Auto` prefers the API runtime over the CLI on Linux, so `Auto` with a `DockerfileImage`, or `OpenReadAsync`, was throwing there.

## Notes for the reviewer

- **No public API change** — everything is `internal`, so `ref/` is untouched.
- Two gaps left in place, both flagged rather than papered over: `ArchiveImage` (`docker load`) is still unsupported by the API runtime, since nothing in the suite covers it and implementing it would mean untested code; and `POST /build` does not send `X-Registry-Config`, so a Dockerfile whose base image lives in a private registry will not authenticate through this path.
- `.dockerignore` is not applied to the build context. It is a client-side filter the daemon never sees, so the whole context directory is sent. This is noted in a comment in `DockerApiTarArchive`.

## Verification

Run locally on arm64 macOS against Docker Desktop and Apple `container`:

- `DockerApiContainerTests`: **22/22** (11 tests x 2 TFMs).
- `AppleContainerContainerTests`: **22/22**.
- Full `Meziantou.Framework.TemporaryContainers.Tests` suite: 212 passed, 2 failed, 29 skipped. The 2 failures are the `SqlServerContainerTests` connection-string tests, the known local baseline on this machine (the mssql image has no arm64 variant); that same test also hung the net11.0 host at the 10-minute dump timeout, so the two classes above were re-run separately to cover both TFMs. Podman and wslc skip locally, neither is installed.
- `dotnet run ./eng/update-all.cs`: no changes.
- `dotnet build -c Release /p:IsOfficialBuild=true /p:TreatsWarningsAsErrors=true`: 0 warnings, 0 errors.

Podman and wslc could not be exercised locally; CI is the first place those gates are tested.

### DIFF
--- a/tests/Meziantou.Framework.TemporaryContainers.Tests/DockerApiContainerTests.cs
+++ b/tests/Meziantou.Framework.TemporaryContainers.Tests/DockerApiContainerTests.cs
@@ -1,0 +1,23 @@
+using Meziantou.Framework.TemporaryContainers.Internals;
+using Meziantou.Xunit;
+
+namespace Meziantou.Framework.TemporaryContainers.Tests;
+
+// Each test starts its own container. Running them all at once saturates the CI agents and makes the container
+// runtimes fail transiently (image pull races, port collisions), so this class does not run in parallel.
+[TestClass(DisableParallelization = true)]
+public sealed class DockerApiContainerTests() : ContainerRuntimeTestsBase(DockerApi)
+{
+    /// <summary>The runtime is not exposed by <see cref="ContainerRuntime"/>, and it caches the connection it probes, so every test shares a single instance instead of reconnecting.</summary>
+    private static readonly ContainerRuntime DockerApi = new DockerApiRuntime();
+
+    // The Linux CI agents ship the Docker daemon, so a socket that does not answer there means the CI setup regressed. Skipping would hide it.
+    // The Windows agents are left out on purpose: their Docker Engine is flaky enough that requiring it would fail runs for reasons unrelated to the change under test.
+    protected override bool IsRuntimeRequired => OperatingSystem.IsLinux() && TestEnvironment.IsOnGitHubActions();
+
+    [Fact]
+    public Task PauseAndUnpause() => AssertPauseUnpauseAsync();
+
+    [Fact]
+    public Task StartAsync_ContainerExitsBeforeTheReadyMessage_ReportsWhatTheContainerPrinted() => AssertStartFailureReportsContainerOutputAsync();
+}
--- a/tests/Meziantou.Framework.TemporaryContainers.Tests/DockerApiContainerTests.cs
+++ b/tests/Meziantou.Framework.TemporaryContainers.Tests/DockerApiContainerTests.cs
@@ -1,4 +1,3 @@
-using Meziantou.Framework.TemporaryContainers.Internals;
 using Meziantou.Xunit;
 
 namespace Meziantou.Framework.TemporaryContainers.Tests;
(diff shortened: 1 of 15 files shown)


tests/Meziantou.Framework.TemporaryContainers.Tests/DockerApiContainerTests.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Meziantou.Framework.TemporaryContainers.Tests;
+
+// Each test starts its own container. Running them all at once saturates the CI agents and makes the container
+// runtimes fail transiently (image pull races, port collisions), so this class does not run in parallel.
+[TestClass(DisableParallelization = true)]
+public sealed class DockerApiContainerTests() : ContainerRuntimeTestsBase(DockerApi)
+{
+    /// <summary>The runtime is not exposed by <see cref="ContainerRuntime"/>, and it caches the connection it probes, so every test shares a single instance instead of reconnecting.</summary>
+    private static readonly ContainerRuntime DockerApi = new DockerApiRuntime();
```
meziantou: I think ContainerRuntime should expose it

meziantou: Done in 1f8d1f761 — `ContainerRuntime.DockerApi` is now public, alongside `Docker`, `Podman`, `AppleContainer` and `Wslc`.

`AutoContainerRuntime` also drops its own private instance and uses the exposed one, so the connection it caches is shared with callers that ask for the runtime explicitly. `ref/Meziantou.Framework.TemporaryContainers.cs` is regenerated, and the readme and package description now mention the Docker Engine API alongside the CLIs.

PR #3016, opened 2026-09-08
Add a CLI that dumps a syntax tree as JSON, and align the four Language libraries

meziantou:
Adds a CLI that dumps the syntax tree of any `Meziantou.Framework.Language.*` document as JSON. Getting there surfaced two inconsistencies between the four libraries, fixed first so the tool has a uniform surface to serialize.

## 1. `RegexFlavor` → `RegexDialect` (`Language.Regex` 1.0.1 → 2.0.0)

`Language.Shell` calls a sub-variant a *dialect*, `Language.Regex` called it a *flavor*. Anything reading both carries two names for one concept.

`RegexFlavor`, `RegexFlavorFamily` and `RegexFlavorFeatures` become `RegexDialect`, `RegexDialectFamily` and `RegexDialectFeatures`; `RegexParseOptions.Flavor`, `RegexSyntaxTree.Flavor` and `RegexSyntaxNode.Flavor` become `Dialect`. The static instances, their `Name` values (`net`, `javascript`, `pcre`, `ere`, `bre`) and every `TryParse` alias are unchanged, so `RegexDialect.TryParse("dotnet")` still resolves as before.

Breaking, so a major bump, with no `[Obsolete]` shims — matching acc74bcd7. `Meziantou.Framework.TemporaryContainers`' unrelated Docker `Flavor` is untouched.

## 2. Absolute source positions in `Language.Xml` (2.0.2 → 3.0.0)

XML was the only one of the four reporting no positions: `XmlSyntaxNode.Span` was `new(0, ToFullString().Length)`, `XmlSyntaxToken` had no `FullSpan`, and `XmlSyntaxTrivia` had no span at all. Only `XmlDiagnostic.Span` carried real offsets, so a diagnostic could not be tied back to the node it reported on.

A `fullStart` is threaded through the base types, the eleven concrete node types and the parser — which already held the offset in a local at every construction site. Because XML nodes build their tokens inline from a substring of their own text, each node also positions its tokens: element and end-tag names via the existing name-start helpers, comment and CDATA text after their fixed delimiters, and the declaration's pseudo-attributes via `ParseAttributeSegments`.

Three points for a reviewer:

- **`XmlSyntaxNode.Span` returns `FullSpan`** rather than copying the JSON library's min/max over descendant tokens. An element's first token is its *name*, so that formula would report the element as starting after the `<`. The two spans differ elsewhere only to exclude trivia, and the XML parser attaches none to nodes. It also stops `Span` being O(subtree).
- **`DescendantTokens()` / `DescendantNodesAndTokens()` were fixed** to yield the node's own tokens; both only walked the children's, so a leaf's token was invisible and `DescendantTrivia()` inherited the hole.
- **The editing paths are deliberately untouched.** `ReplaceNode` / `ReplaceToken` / `ReplaceTrivia` keep searching the text, because `SyntaxFactory`-built trivia is anchored at 0 — collapsing them onto the new spans would regress `ReplaceTrivia`. That simplification is worth a separate change. All 40 pre-existing XML tests passing unchanged is the evidence nothing regressed.

Still out of scope: the XML parser produces **no trivia at all** (`new XmlSyntaxTrivia` appears nowhere in the library; whitespace stays inside the node's text). `XmlSyntaxTrivia.Span` is added for API parity but reads 0 on a parsed tree. Teaching the parser to lex trivia is a redesign of the text-based node model.

## 3. `Meziantou.Framework.Language.Tool` (new, 1.0.0)

```bash
meziantou.language syntax --input app.csproj
echo '{"a":1}' | meziantou.language syntax --language json
```

`syntax` is the only command today; the tool is laid out this way so others can join it.

**`--language`** takes a flat, fine-grained value — `json`, `xml`, `regex[-dialect]`, or a shell dialect. The suffixes go through the libraries' own `TryParse`, so their aliases (`regex-dotnet`, `regex-js`, `pwsh`, `batch`, …) come for free with no new table. Resolution is json/xml → `regex-`-prefixed → shell, which keeps the `posix` alias the two share unambiguous. Omitted, the language is detected from the extension of `--input`, which is why it is required for stdin; regular expressions are never detected, having no file form.

**The document is the same shape for every language**, built only from base-class members, so no node type needs its own code and new ones need no change. `--no-tokens`, `--no-trivia`, `--no-text` and `--no-spans` slim it down. `"language"` is the library family and the sub-variant sits in `"dialect"`, so regex does not encode its dialect twice.

`childNodes` and `tokens` are separate arrays rather than one in source order, because `ChildNodesAndTokens()` does not agree across the four libraries — regex and shell merge by position, JSON and XML concatenate children then tokens. One merged array would silently serialize the same structure differently per language.

Three things a reviewer might otherwise wonder about:

- **The walk uses an explicit stack.** Regex and shell cap their parse depth; JSON and XML do not, so a recursive writer would overflow on input that parsed perfectly well.
- **`Utf8JsonWriter`'s depth limit is raised** — a node costs about three JSON levels, so the default 1000 is reached by a ~350-deep input.
- **The relaxed encoder is used**, because the default escapes `<`, `>`, `&` and every non-ASCII character, which makes an XML dump unreadable.

One writer file per language is mandatory, not stylistic: all four namespaces export `TextSpan`, `SourceText`, `TextLine` and `SyntaxFactory`, so two `using Meziantou.Framework.Language.*` in one file naming any of them is CS0104. The field names and shapes live in a single shared `SyntaxTreeJsonWriter` so the four dumps cannot drift.

## Verification

`dotnet build eng/build.proj /p:TreatWarningsAsErrors=true` — succeeded, 0 warnings.

| Suite | Tests | Failed |
| --- | --- | --- |
| `Language.Regex` | 1140 | 0 |
| `Language.Shell` | 2898 | 0 |
| `Language.Xml` | 112 (40 pre-existing + 72 new) | 0 |
| `Language.Json` | 46 | 0 |
| `Language.Tool` | 100 | 0 |

Both `ref/` files regenerated with `-c Release /p:IsOfficialBuild=true` and stable across a second run: the regex diff is 20 lines, all renames; the XML diff only the added parameters and the new `XmlSyntaxToken.FullSpan` / `XmlSyntaxTrivia.Span` / `FullSpan`.

`dotnet run ./eng/update-all.cs` was run until it reported no changes. It produced `slnx/Meziantou.Framework.Language.Tool.slnx`, the solution and root-README entries, and the `<!-- help -->` block including a `### syntax` section.

The tool was also driven end to end over all four languages, checking that auto-detection picks the right one, that the output is a single line, that each `--no-*` flag removes its fields, that malformed input still exits 0 with a populated `diagnostics` array, and that `text.Substring(fullSpan)` slices back to every node's own text.


meziantou:
Superseded by a stack: #3017 (regex dialect rename) → #3018 (XML source positions) → #3019 (the CLI). Same content, split so each layer can be reviewed and merged on its own.

PR #3019, opened 2026-09-08
Add a CLI that dumps a syntax tree as JSON

meziantou:
**Stack 3 of 3** — based on #3018, which is based on #3017. This is the change the other two were groundwork for.

None of the four `Language.*` libraries could be inspected from outside .NET: no serialization, no debug dump, and only `ToFullString()`, which hands the source text back. Debugging a parse, diffing two trees, or feeding a tree to a script meant writing a throwaway program.

```bash
dotnet tool install --global Meziantou.Framework.Language.Tool
Meziantou.Framework.Language.Tool syntax --input app.csproj
echo '{"a":1}' | Meziantou.Framework.Language.Tool syntax --language json
```

`syntax` is the only command today; the tool is laid out with a command level so others can join it later.

## `--language`

A flat, fine-grained value — `json`, `xml`, `regex[-dialect]`, or a shell dialect. The suffixes go through the libraries' own `TryParse`, so their aliases (`regex-dotnet`, `regex-js`, `regex-grep`, `pwsh`, `batch`, …) come for free with no new table to maintain. Resolution runs json/xml → `regex-`-prefixed → shell, which keeps the `posix` alias the two libraries share unambiguous: bare `posix` is the shell, `regex-posix` is POSIX ERE.

Omitted, the language is detected from the extension of `--input` — which is why it is required when reading stdin. Regular expressions are never detected, having no file form.

## Output

Non-indented, one line, UTF-8 without BOM. The document is the same shape for every language, built **only from base-class members**, so no node type needs its own code and new ones need no change:

```jsonc
{
  "language": "shell",   // json | xml | regex | shell
  "dialect": "bash",     // regex and shell only
  "diagnostics": [ { "id": "…", "severity": "Error", "message": "…", "span": { "start": 3, "length": 1 } } ],
  "root": {
    "kind": "Command", "type": "ShellCommandSyntax",
    "span": { "start": 0, "length": 7 }, "fullSpan": { "start": 0, "length": 8 },
    "text": "echo hi",
    "tokens": [ { "kind": "IdentifierToken", "text": "echo", "leadingTrivia": [], "trailingTrivia": [] } ],
    "childNodes": [ ]
  }
}
```

Empty collections and default values are left out, and `valueText` appears only when it differs from `text`. `--no-tokens`, `--no-trivia`, `--no-text` and `--no-spans` slim it down.

`"language"` is the library family and the sub-variant sits in `"dialect"`, so regex does not have to encode its dialect twice — this is the shape #3017 made possible.

A parse problem lands in `diagnostics` and does **not** fail the run: the four parsers never throw, and a tree is produced for invalid input too. The exit code is non-zero only for a usage or I/O problem.

## Decisions a reviewer might question

- **`childNodes` and `tokens` are separate arrays**, not one in source order. `ChildNodesAndTokens()` does not agree across the four libraries — regex and shell merge by `FullSpan.Start`, JSON and XML concatenate children then tokens — so one merged array would silently serialize the same structure differently per language.
- **The walk uses an explicit stack, not recursion.** Regex and shell cap their parse depth (`MaxRecursionDepth`), but JSON and XML have no guard, so a recursive writer would overflow on input that parsed perfectly well. The libraries made `DescendantNodes()` iterative for the same reason.
- **`Utf8JsonWriter`'s depth limit is raised.** A node costs about three JSON levels, so the default 1000 is reached by a ~350-deep input.
- **The relaxed encoder is used**, because the default escapes `<`, `>`, `&` and every non-ASCII character — which turns an XML dump into `"<root>"`.
- **One writer file per language is mandatory, not stylistic.** All four namespaces export `TextSpan`, `SourceText`, `TextLine` and `SyntaxFactory`, so two `using Meziantou.Framework.Language.*` in one file naming any of them is CS0104. The field names and shapes live in a single shared `SyntaxTreeJsonWriter` so the four dumps cannot drift.

No `ToolCommandName` and no custom `RootNamespace`: the tool is invoked by its package name, like six of the eight tools in `src/`.

## Verification

`dotnet build eng/build.proj /p:TreatWarningsAsErrors=true` succeeded with 0 warnings. On this branch: `Language.Tool` **100**, `Language.Xml` **112**, `Language.Json` **46**, `Language.Shell` **2898** — **0 failed**.

The tests drive `Program.MainImpl` in process through `common/ConsoleHelper.cs`, never spawning the exe. They cover auto-detection for thirteen extensions, every regex and shell dialect, stdin with and without `--language`, a missing file, an unknown extension, an unknown language, `--output` writing without a BOM, single-line output, each `--no-*` flag removing exactly its fields, malformed input still exiting 0 with populated `diagnostics`, and — across all four languages — that `text.Substring(fullSpan)` slices back to every node's own `text`, which is the cross-check that #3018 actually landed.

`dotnet run ./eng/update-all.cs` was run until it reported no changes. It produced `slnx/Meziantou.Framework.Language.Tool.slnx`, the solution and root-`README.md` entries, and the `<!-- help -->` block including its `### syntax` section.


meziantou:
Superseded by #3071, rebuilt on the shared green/red tree from #3052 and based directly on main. The four per-language writers collapsed into one, and `childNodes`/`tokens` merged into a single ordered `children` array now that ChildNodesAndTokens() is uniform.

PR #3052, opened 2026-09-09
Move the four language packages onto a shared green/red syntax tree

meziantou:
Moves all four `Meziantou.Framework.Language.*` languages onto a shared Roslyn-style green/red syntax tree, and puts the tree machinery in the common `Meziantou.Framework.Language` project.

## Why

Every language had its own red-only tree, and every one of them had the same three problems.

- **Each node stored its own copy of the source text** and an absolute span, so a node was bound to one position in one document and the text was duplicated once per nesting level.
- **Every edit was a lie.** `ReplaceNode`, `ReplaceToken` and `ReplaceTrivia` spliced text and re-parsed, returned the wrong node type, and fell back to a substring search that *silently did nothing* when the text occurred twice.
- **Four near-identical copies** of `*SyntaxNode`, `*SyntaxToken`, `*SyntaxTrivia`, `*SyntaxNodeOrToken`, `*SyntaxVisitor`, `*SyntaxRewriter` — already drifting apart.

Roslyn solves exactly this with a green/red split: internal, position-independent, shareable green nodes storing widths, and lazily-materialised red nodes adding a parent pointer and an absolute position. That split is what makes typed `WithX`/`ReplaceNode`, annotations, and incremental reparse possible at all.

## What changed

`Meziantou.Framework.Language` grew from 320 to ~4,700 lines and now holds the whole tree: `GreenNode` with node flags OR'd upward, green tokens/trivia/lists, the red `SyntaxNode`/`SyntaxToken`/`SyntaxTrivia`/`SyntaxNodeOrToken` and the six list types, `ChildSyntaxList`, `SyntaxTree`, `SyntaxAnnotation`, `SyntaxWalker`, navigation, `SyntaxReplacer`, `SyntaxDiffer`, and `RemoveNode`/`SyntaxRemoveOptions`. It also gained the text primitives the tree needs — set operations on `TextSpan`, a binary-search `SourceText.GetLine`, `GetSubText`, `GetChangeRanges`, `TextChangeRange`, `Diagnostic.Create`.

Then each language moved onto it:

| | |
| --- | --- |
| **JSON** 3.0.0 → **4.0.0** | separated lists (`JsonArrayElementSyntax` deleted), a rewritten parser with context-sensitive follow sets, trailing trivia, and **incremental reparse** — `WithChangedText` reuses unchanged green subtrees |
| **Regex** 3.0.0 → **4.0.0** | 37 node types generated by a `Syntax.tt`; per-node options became green constructor state |
| **XML** 3.0.0 → **4.0.0** | needed a **real tokenizer** first: the parser never tokenized `<`, `>`, `/>`, `=` or the quotes, so a node's text could not be its slots |
| **Shell** 2.0.0 → **3.0.0** | the largest: 89 node types across three dialects, ~7,000 lines of parser, 511 lines of template for 11,087 generated |
| **Language** 1.0.0 → **1.1.0** | additive only |

## Things worth a reviewer's attention

**One deliberate departure from Roslyn.** `GreenNode.WithSlots(ReadOnlySpan<GreenNode?>)` — one mechanical override per green type — lets the whole replace engine live in the common layer instead of being written per language. Saves roughly 400 lines each; the public shape is unchanged. A node that reorders or drops slots in its override would silently corrupt every edit, so there is a generic round-trip test over a corpus.

**Three design calls that were not obvious:**

- *Regex per-node options* (`Options`, `InnerOptions`, `AppliedOptions`) became green constructor state rather than a positional map. Two textually identical subtrees under different inline options genuinely differ and must not be shared.
- *XML whitespace* is trivia inside a tag and character data between tags, because that is what XML says. This is why `Span` and `FullSpan` can now differ in XML, which they never did before. A self-closing element also became its own node type, matching the two kinds the enum already reported.
- *The shell here-document back-link* was a mutable pointer the parser set after the fact — impossible for a shared, position-independent green node. It is derived from the tree now: bodies after a statement pair up in order with the `<<` redirections inside it.

**Bugs found along the way, not by the test updates:**

- The shared replacer could not reach a token inside a token-list slot, because such a list has no red node and the walk only descended into slots that did. Replacing a token of a skipped-text node found nothing and threw. Found by the shell suite; it affected JSON and regex too.
- `IsEquivalentTo` ignored trivia in name only — it compared `FullWidth`.
- A missing token built with trivia reported `IsMissing` false, because trivia passes `IsNotMissing` up to its token.
- `SourceText.GetChangeRanges` could cut a surrogate pair in half.
- A round-trip theory over every prefix and single-character deletion of the XML corpus found an end tag at end-of-document losing the whitespace before its missing `>`.

## Breaking changes

Every language: `tree.Root`/`Text`/`Diagnostics` → `GetRoot()`/`GetText()`/`GetDiagnostics()`; `XSyntaxKind` → `SyntaxKind`; `node.Kind` → `node.Kind()`; the per-language token/trivia/node-or-token classes → the shared structs; public node constructors → `SyntaxFactory`; rewriter overrides return `SyntaxNode?`; `DefaultVisit` no longer recurses (a `*SyntaxWalker` does).

Two behavioural changes are deliberate and easy to miss: **`ReplaceNode` no longer carries the replaced node's trivia over** (ask with `WithTriviaFrom`), and it **throws rather than silently doing nothing** for a node from another tree. Each readme has a migration table.

## Verification

- 13 Release builds clean with analyzers and `PublicApiGeneratorVerifyNoChangeOnBuild`.
- Tests, all TFMs, 0 failures: Language 638, Json 180, Regex 1142, Shell 2974, Xml 230, **DependencyScanning 236**.
- The dependency-scanning suite is the real regression guard — it drives the JSON and XML trees against real project files — and it passes **unedited**.
- Json, Regex and Shell also pass the `InvariantGlobalization=true` CI leg.
- `dotnet run ./eng/update-all.cs` leaves the tree clean; `ref/` is regenerated and committed.

Every readme sample was executed against the built package rather than written from the API; most had drifted.

## Not in scope

Regex and shell keep diagnostics as a flat list on the tree rather than on green nodes — threading node-relative offsets through seven parsers is a separate change. Incremental reparse is JSON only.


src/Meziantou.Framework.Language.Xml/SyntaxFacts.cs
```diff
@@ -0,0 +1,33 @@
...
+        SyntaxKind.EqualsToken => "=",
+        SyntaxKind.SingleQuoteToken => "'",
+        SyntaxKind.DoubleQuoteToken => "\"",
+        SyntaxKind.LessThanQuestionToken => "<?",
+        SyntaxKind.QuestionGreaterThanToken => "?>",
+        SyntaxKind.XmlCommentStartToken => "<!--",
+        SyntaxKind.XmlCommentEndToken => "-->",
+        SyntaxKind.CDataStartToken => "<![CDATA[",
+        SyntaxKind.CDataEndToken => "]]>",
+        _ => "",
```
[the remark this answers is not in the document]
meziantou: Fixed: `GetText` now returns `<?xml` and `<!DOCTYPE` for those kinds, with a remark noting they are the canonical spellings — a document may write either in any case, and the token then carries what the document said.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SourceText.cs
```diff
@@ -49,23 +47,92 @@ public TextLine GetLine(int position)
         ArgumentOutOfRangeException.ThrowIfNegative(position);
```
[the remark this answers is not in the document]
meziantou: Fixed: `GetLine` now rejects a position past `Length`. `Length` itself is still accepted, since that is where a span running to the end of the text finishes. The sweep test that deliberately ran past the end was updated to assert the rejection instead.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/Syntax/SyntaxTreeDiagnostics.cs
```diff
@@ -0,0 +1,91 @@
...
+                {
+                    yield return diagnostic;
+                }
+
+                continue;
+            }
+
+            foreach (var info in current.GetDiagnostics())
+            {
+                yield return info.ToDiagnostic(currentPosition, sourceText);
```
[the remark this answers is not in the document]
meziantou: Fixed: the walk still prunes clean subtrees on a flag, but the results are now ordered by position afterwards. `OrderBy` is stable, so two diagnostics at the same position keep the order the walk found them in.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,569 @@
...
+            }
+
+            kept.Add(items[i]);
+        }
+
+        var residual = SyntaxNodeRemover.ResidualTrivia(removed, options, parent.Green);
+        if (residual is not null && insertionIndex >= 0)
+        {
+            if (insertionIndex < kept.Count)
+            {
```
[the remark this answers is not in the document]
meziantou: Fixed: each contiguous run of removals now keeps its trivia where it was, instead of pooling every removal's trivia and attaching it at the first gap. Added `RemoveNodes_KeepsEachRemovedMembersTriviaWhereItWas`, which fails on the old code and passes on the new.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SyntaxRemoveOptions.cs
```diff
@@ -0,0 +1,29 @@
+namespace Meziantou.Framework.Language;
+
+/// <summary>What to do with the trivia around a node that is being removed.</summary>
+/// <remarks>
+/// A node's leading and trivia would normally disappear with it, which quietly throws away any comment sitting in
```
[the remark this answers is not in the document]
meziantou: Fixed: "leading and trailing trivia".

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language.Xml/SyntaxFacts.cs
```diff
@@ -0,0 +1,33 @@
...
+    };
+
+    /// <summary>Determines whether <paramref name="kind"/> is whitespace or a line break.</summary>
+    public static bool IsTrivia(SyntaxKind kind) => kind is SyntaxKind.WhitespaceTrivia or SyntaxKind.EndOfLineTrivia;
+
+    /// <summary>Determines whether <paramref name="value"/> may begin an XML name.</summary>
+    public static bool IsNameStartCharacter(char value) => char.IsLetter(value) || value is '_' or ':';
+
+    /// <summary>Determines whether <paramref name="value"/> may appear inside an XML name.</summary>
+    public static bool IsNameCharacter(char value) => char.IsLetterOrDigit(value) || value is '_' or ':' or '-' or '.';
```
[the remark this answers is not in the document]
meziantou: Not changed, deliberately. These two predicates are the ones the XML parser has always used — I moved them verbatim from `XmlSyntaxTree.XmlParser` into `SyntaxFacts` — so the limitation predates this PR. Widening what parses (`·`, combining marks, supplementary-plane name characters) changes which documents are accepted and which text becomes skipped text, which is not something to slip into a refactor whose contract is that every existing test keeps passing. Worth its own change.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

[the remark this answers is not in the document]
meziantou: Implemented after all, at the maintainer's request.

`SyntaxFacts` now carries `NameStartChar` and `NameChar` from XML 1.0 verbatim, and both predicates take a `Rune` rather than a `char` so a name character outside the basic plane is a single operand instead of two halves. The parser reads names in scalar values to match, and a lone surrogate — no scalar value at all — ends a name rather than joining it.

It differs from the old letter/digit rule in both directions: `<a·b/>`, combining marks, `‿`, and `<𐌰/>` are now names, while `ª` and `µ` are letters XML does not let a name begin with. The loops that read a name still don't require a name *start* character, so `<1bad>` keeps round-tripping as skipped text.

All eight new assertions fail against the old predicates and pass against the new; the whole corpus round-trips. Commit d4118dd36.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,620 @@
...
+            // A separated list has to keep alternating, so the separator that went with the node goes too: the one
+            // after it, or the one before it when nothing follows. A neighbour that is a node rather than a token is
+            // a sibling in a list that has no separators at all, and taking it would delete something nobody asked to.
+            if (index + 1 < items.Length && items[index + 1] is { IsToken: true } && !removing.Contains(index + 1))
+            {
+                removing.Add(index + 1);
+            }
+            else if (index > 0 && items[index - 1] is { IsToken: true })
+            {
+                removing.Add(index - 1);
```
meziantou: [P2] Remove the separator before a deleted suffix only once per range

When adjacent nodes at the end of a separated list are removed together, both iterations can select the same separator, leaving the separator before the deleted range behind. Reproduced by parsing `a|b|c` with `RegexDialect.Net` and calling `root.RemoveNodes([alternation.Branches[1], alternation.Branches[2]], SyntaxRemoveOptions.KeepNoTrivia)`: the result is `a|`, rather than `a`. This introduces an empty alternative, so the resulting regex now matches inputs that the remaining branch would reject. Compute separator removal over contiguous ranges (including the separator before a removed suffix), rather than independently for each original index.

src/Meziantou.Framework.Language/Syntax/SyntaxReplacer.cs
```diff
@@ -0,0 +1,206 @@
...
+    /// <summary>
+    /// Registers a change to the slots of <paramref name="node"/>, applied once everything below it has been rebuilt.
+    /// </summary>
+    /// <remarks>
+    /// Running after the children means a change to a node and a change to something inside it both take effect: the
+    /// walk still goes all the way down, and each change is applied on the way back out.
+    /// </remarks>
+    public void EditSlots(SyntaxNode node, Func<GreenNode?[], GreenNode?[]> edit)
+    {
+        _slotEdits[node] = edit;
```
meziantou: [P2] Compose slot edits registered for the same parent

`RemoveNodes` groups targets by `(parent, slot)` and calls `EditSlots` once for each group, but this dictionary retains only the last callback for a parent. Consequently, removing children from two slots of the same node silently applies only one removal. For example, parse `if true; then echo a; elif false; then echo b; else echo c; fi` as Bash and remove both `conditional.ElifClauses[0]` and `conditional.ElseClause` in one call: the `elif` clause remains. Compose the callbacks, or group all slot changes for a parent into one callback, so every requested removal takes effect.

src/Meziantou.Framework.Language.Regex/RegexSyntaxTree.cs
```diff
@@ -1,46 +1,53 @@
...
+    public override string? FilePath => null;
+
+    public override SourceText GetText() => _text;
+
+    /// <summary>Gets the root of the tree.</summary>
+    public new RegexPatternSyntax GetRoot() => _root;
 
-    /// <summary>Parses <paramref name="pattern"/> as a complete pattern. Never throws; problems are reported as diagnostics.</summary>
+    /// <summary>Gets everything the parser had to report, in source order.</summary>
+    public new IReadOnlyList<Diagnostic> GetDiagnostics() => _diagnostics;
```
meziantou: [P2] Override diagnostic retrieval instead of hiding it

The new shared `SyntaxTree.GetDiagnostics()` is virtual, but `new` bypasses this implementation whenever the tree is accessed through that base type, including through `node.SyntaxTree`. Reproduced with `var tree = RegexSyntaxTree.ParseText("(", RegexDialect.Net)`: `tree.GetDiagnostics().Count` is 1, while `((SyntaxTree)tree).GetDiagnostics().Count()` is 0. XML and Shell use the same hiding pattern; `<a>` likewise loses its XML error through the base API. Override the virtual method in the trees that store flat diagnostics so language-independent consumers still receive parser errors; this does not require moving diagnostics onto green nodes.

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,620 @@
...
+    public static TRoot RemoveNodes<TRoot>(this TRoot root, IEnumerable<SyntaxNode> nodes, SyntaxRemoveOptions options)
+        where TRoot : SyntaxNode
+    {
+        ArgumentNullException.ThrowIfNull(root);
+        ArgumentNullException.ThrowIfNull(nodes);
+
+        var targets = nodes.ToArray();
+        foreach (var target in targets)
+        {
+            ArgumentNullException.ThrowIfNull(target, nameof(nodes));
```
meziantou, line 167: [P2] Validate every removal target against the supplied root

Unlike `ReplaceNodes`, this loop checks only for null, and `TryLocate` later establishes membership in the target's own parent rather than in `root`. A batch containing a valid target and a node from another tree therefore succeeds partially despite the documented `ArgumentException` contract. Reproduced with arrays parsed from `[1,2]` and `[3,4]`: `first.RemoveNodes([first.Elements[0], second.Elements[0]], SyntaxRemoveOptions.KeepNoTrivia)` returns `[2]` instead of rejecting the foreign target. Call `EnsureInTree` for every target before ancestor filtering and registration.

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,620 @@
...
+        result.InsertRange(index + 1, [separator, .. Interleave(replacements, separator, separatorAtEnd: false)]);
+
+        return [.. result];
+    }
+
+    private static GreenNode? FindSeparator(GreenNode?[] items) => Array.Find(items, item => item is { IsToken: true });
+
+    private static List<GreenNode?> Interleave(GreenNode?[] nodes, GreenNode separator, bool separatorAtEnd)
+    {
+        var result = new List<GreenNode?>((nodes.Length * 2) - 1);
```
meziantou, line 551: [P2] Treat insertion of an empty node sequence as a no-op

For zero replacements this capacity is -1, so inserting an empty enumerable into a separated list throws `ArgumentOutOfRangeException` from the `List` constructor. Reproduced with an array parsed from `[1,2]` and `array.InsertNodesAfter(array.Elements[0], Array.Empty<SyntaxNode>())`; `InsertNodesBefore` fails too. This also affects callers whose filtered insertion sequence happens to be empty. Return the original tree before splicing when an insertion has no replacements; merely clamping the capacity is insufficient because the insert-after branch would still add a separator.

meziantou (commented):
Reviewed d3d651dbcd525e69caf35ec4dac301c8fe8a48b8. Five actionable findings below, each reproduced against the built PR assemblies.

Validation: Release builds and tests for Language, JSON, Regex, Shell, XML, and DependencyScanning on both net10.0 and net11.0, using `/p:TreatsWarningsAsErrors=true`: 5,518 passed, 0 failed, 0 skipped. `dotnet run ./eng/update-all.cs /p:TreatsWarningsAsErrors=true` completed successfully and left the worktree clean. The existing suites pass, but the reproductions expose the cases described in the inline comments.

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
meziantou: Fixed: the separators are chosen per run of adjacent removals now, rather than independently per node — two neighbours could otherwise settle on the same separator. `a|b|c` minus branches 1 and 2 gives `a`; minus 0 and 1 gives `c`. Both directions are tested.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/Syntax/SyntaxReplacer.cs
meziantou: Fixed: `EditSlots` composes with any callback already registered for that node instead of replacing it. Your `elif`/`else` case is now a test.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language.Regex/RegexSyntaxTree.cs
meziantou: Fixed: the shared `GetDiagnostics()` returns `IReadOnlyList<Diagnostic>` so the four trees can `override` rather than hide it, which they now do. JSON still materialises the walk over its green nodes; Regex, XML and Shell hand back their flat list. Tested through the base type.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
meziantou: Fixed: `RemoveNodes` calls `EnsureInTree` for every target before the ancestor filter, so a foreign node is rejected rather than partially applied. Your `[1,2]`/`[3,4]` case is now a test.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
meziantou: Fixed: an insertion with no replacements returns the tree unchanged, before any splicing — so the insert-after branch never gets to add its separator either. Only a replacement may be empty, and that one still means "take the node out".

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

meziantou (commented):
Reviewed 0799030db854b7897f46b3929ccbb4c1d110b6bf. Four additional actionable findings below, reproduced against the built PR assemblies.

Validation: Release builds and tests for Language, JSON, Regex, Shell, XML, and DependencyScanning on net10.0 and net11.0 using `/p:TreatsWarningsAsErrors=true`: 5,530 passed, 0 failed, 0 skipped. `dotnet run ./eng/update-all.cs /p:TreatsWarningsAsErrors=true` completed successfully and left the worktree clean. The existing suites pass; the inline comments describe additional cases they do not exercise.

src/Meziantou.Framework.Language/SeparatedSyntaxList.cs
```diff
@@ -0,0 +1,288 @@
...
+    public SeparatedSyntaxList<TNode> Insert(int index, TNode node)
+    {
+        ArgumentNullException.ThrowIfNull(node);
+        ArgumentOutOfRangeException.ThrowIfNegative(index);
+        ArgumentOutOfRangeException.ThrowIfGreaterThan(index, Count, nameof(index));
+
+        if (Count == 0)
+            return new SeparatedSyntaxList<TNode>(new SyntaxNodeOrTokenList([node]));
+
+        var separator = CreateSeparator(node);
```
meziantou, line 155: [P2] Choose separators from the list's grammar context

`CreateSeparator(node)` delegates to the inserted node's language-wide default, which is `;` for every shell node. For a pipeline parsed from `a | b`, `pipeline.WithCommands(pipeline.Commands.Add(SyntaxFactory.Command(ShellDialect.Bash, "c"))).ToFullString()` produces `a | b;c`: the added command is no longer part of the pipeline. The shared `InsertNodesAfter` path has the same problem when a pipeline contains only one command and has no existing separator to copy (`SyntaxFactory.Pipeline(commandA)` becomes `a;b`). Preserve or resolve the separator for the owning list/slot, including singleton lists, rather than using the element's language alone; shell lists can require pipes, logical operators, or commas as well as semicolons.

src/Meziantou.Framework.Language.Shell/ShellRedirectionSyntax.cs
```diff
@@ -1,54 +1,67 @@
...
+    /// <remarks>
+    /// Counted over the whole statement the bodies are ordered against, not over the nearest statement above this
+    /// redirection. A pipeline is one statement holding several, and its bodies follow it in one sequence, so
+    /// counting inside the nearer one would give every branch of the pipeline the same ordinal.
+    /// </remarks>
+    private int CountPrecedingHereDocumentsIn(ShellStatementSyntax owner)
     {
-        ArgumentNullException.ThrowIfNull(operatorToken);
+        var ordinal = 0;
+        foreach (var redirection in owner.DescendantNodes().OfType<ShellRedirectionSyntax>())
```
meziantou: [P2] Exclude nested statement lists when pairing here-documents

This traversal counts redirections inside nested process/command substitutions even though their bodies belong to a different statement list. Reproduced with Bash text `cat <(cat <<INNER\ninner\nINNER\n) <<OUTER\nouter\nOUTER\n`: parsing round-trips with zero diagnostics, but the `<<OUTER` redirection's `HereDocument` is null, and the outer body's `Redirection` incorrectly points to `<<INNER`. Only count redirections whose bodies belong to the current statement list, while retaining traversal through pipeline branches; apply the same scope restriction to the reverse lookup in `PosixHereDocumentSyntax.Redirection`.

src/Meziantou.Framework.Language/SyntaxNode.cs
```diff
@@ -0,0 +1,712 @@
...
+    /// <param name="getInnermostNodeForTie">
+    /// When several nested nodes cover exactly the same range, whether to return the innermost rather than the outermost.
+    /// </param>
+    /// <exception cref="ArgumentOutOfRangeException"><paramref name="span"/> is not inside this node.</exception>
+    public SyntaxNode FindNode(TextSpan span, bool getInnermostNodeForTie = false)
+    {
+        if (!FullSpan.Contains(span))
+            throw new ArgumentOutOfRangeException(nameof(span));
+
+        var node = FindToken(span.Start).Parent ?? this;
```
meziantou, line 404: [P2] Honor the outermost-node option for equal spans

The search starts at the token's immediate parent and stops as soon as that node contains the span. With `getInnermostNodeForTie: false`, there is no subsequent ascent through ancestors with the same full span, so it still returns the innermost match. For `var root = JsonSyntaxTree.ParseText("1").GetRoot()`, both `root.FindNode(root.FullSpan)` and `root.FindNode(root.FullSpan, true)` return `JsonNumberSyntax`, although the default option promises the outermost match (`JsonDocumentSyntax` here). Ascend through equal-span ancestors, bounded by the receiver, when the option is false so callers can reliably select the intended enclosing construct.

src/Meziantou.Framework.Language.Json/JsonSyntaxRewriter.cs
```diff
@@ -1,97 +1,145 @@
...
-    protected virtual JsonSyntaxNode? VisitCore(JsonSyntaxNode node)
+    public override SyntaxNode? VisitJsonSkippedText(JsonSkippedTextSyntax node)
     {
         ArgumentNullException.ThrowIfNull(node);
 
-        return Visit(node);
+        return node.Update(VisitList(node.Tokens));
+    }
+
+    public virtual SyntaxToken VisitToken(SyntaxToken token) => token;
```
meziantou: [P2] Dispatch token trivia through the rewriter's VisitTrivia hook

`VisitToken` returns the token directly, and no traversal calls `VisitTrivia`. Consequently a subclass that overrides `VisitTrivia` to remove comments or rewrite whitespace silently does nothing: visiting the root parsed from `/*hello*/1` invokes the override zero times and leaves the comment intact. Regex, Shell, and XML have the same implementation. Have the default token visit rewrite both trivia lists through the hook and rebuild the token when they change, so the newly exposed trivia override participates in ordinary tree rewriting.

src/Meziantou.Framework.Language/SeparatedSyntaxList.cs
meziantou: The separator now comes from the list when it already holds one, and otherwise from the slot that owns it -- each of the eleven separated lists in the shell grammar declares the separator it holds, so `pipeline.Commands.Add(...)` produces `a | b | c` rather than `a | b;c`. Test: `AddingToAPipelineKeepsItAPipeline`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language.Shell/ShellRedirectionSyntax.cs
meziantou: The walk now stops at a nested statement list, so a redirection inside a substitution no longer counts toward the outer pairing (it still crosses the branches of a pipeline, which is one statement). Applied to both directions of the link, via a shared `HereDocumentRedirectionsIn`. Test: `HereDocumentsInsideASubstitutionDoNotShiftTheOnesOutside`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SyntaxNode.cs
meziantou: The walk up stopped at the first node containing the span and never looked at ancestors covering the same range. It now keeps ascending while the parent spans exactly the same text, bounded by the node being asked, so `getInnermostNodeForTie: false` returns the outermost of the tie. Test: `FindNode_ReturnsTheOutermostNodeSharingTheSpanUnlessAskedForTheInnermost`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language.Json/JsonSyntaxRewriter.cs
meziantou: `VisitToken` now puts each token's trivia through `VisitTrivia` (via a new `VisitList(SyntaxTriviaList)`), in all four languages, and returning the default trivium removes it -- which needed `SyntaxTriviaList` to be constructible from trivia, as `SyntaxTokenList` already was. Test: `Rewriter_PutsTheTriviaOfEveryTokenThroughVisitTrivia`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,643 @@
...
+            {
+                result.RemoveAt(index);
+            }
+
+            result.InsertRange(removeOriginal || insertBefore ? index : index + 1, replacements);
+
+            return [.. result];
+        }
+
+        var separator = FindSeparator(items) ?? original.CreateSeparator()
```
meziantou: `Splice` falls back to the language-wide default separator instead of the one this slot declares, so `InsertNodesBefore` / `InsertNodesAfter` / `ReplaceNode(node, nodes)` insert the wrong token whenever the list does not hold a separator yet.

`FindSeparator(items)` returns `null` for a list of one element — `items` holds nothing but the element — and the fallback is `original.CreateSeparator()`, the *element's* language-wide default, rather than `parent.Green.CreateSeparator(slot)`. That slot-aware overload is exactly what `Syntax.tt` generates the `CreateSeparator(int)` overrides for, and `SeparatedSyntaxList.CreateSeparator` (SeparatedSyntaxList.cs:194) already asks it; this path is the only caller that doesn't, which makes those generated overrides dead here.

Scenario: `ShellPipelineSyntax` declares `"s:Commands:ShellStatementSyntax:PipeToken"`. Parse `a`, take the pipeline's single command and call `root.InsertNodesAfter(command, [otherCommand])`. No token is in `items`, so `ShellSyntaxNode.CreateSeparator()` answers with `SemicolonToken` and the result is a `Pipeline` node whose text is `a;b`. Same for `PowerShellArrayLiteralSyntax` (comma), `PosixCaseClauseSyntax.Patterns` (pipe) and `ShellCommandListSyntax` (`&&`).

`parent` is in scope at the call site (line 469), so the slot's separator can be resolved there and passed in.

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,643 @@
...
+            // Inserting nothing is not an error: it is a caller whose sequence happened to come out empty. Only a
+            // replacement may be empty, and that one means "take the node out".
+            return root;
+        }
+
+        var slotGreen = parent.Green.GetSlot(slot);
+
+        GreenNode?[] items;
+        int start;
+        if (indexInList < 0)
```
meziantou: `TryLocate` reports `indexInList = -1` for two different situations, and this branch treats them alike:

1. the node fills a single-node slot (`n:` / `n?:`), and
2. the node is the only element of a list slot — because a green list of one element collapses to the element itself, so `childGreen.IsList` is `false` and `TryLocate` takes the non-list branch.

For case 2 the code below is right. For case 1 line 477 writes `SyntaxList.ListNode(spliced)` into a slot that expects a typed child, instead of the `ArgumentException` documented on `InsertNodesBefore` / `InsertNodesAfter` / `ReplaceNode(node, IEnumerable<SyntaxNode>)` ("is not somewhere several nodes can go"). The generated `WithSlots` overrides take untyped `GreenNode` and don't check, so the corrupt tree is built silently.

Scenario:

```csharp
var tree = JsonSyntaxTree.ParseText("""{"a":1}""");
var member = ((JsonObjectSyntax)tree.GetRoot().Value!).GetMember("a")!;
var updated = tree.GetRoot().InsertNodesAfter(member.Value, [SyntaxFactory.JsonNumber("2")]);
```

`updated.ToFullString()` is `{"a":12}`, and reading the member's `Value` throws `InvalidCastException`: `JsonMemberSyntax.Value` is `GetRed(ref _value, 2)`, which casts the slot's red projection to `JsonValueSyntax`, but the slot now holds a `SyntaxListNode`.

`IsSeparatedListSlot` can't disambiguate this, since a plain `l:` slot answers `false` too — telling the two apart needs a general "does this slot hold a list" question on the green node.

src/Meziantou.Framework.Language/Syntax/SyntaxReplacer.cs
```diff
@@ -0,0 +1,208 @@
...
+        GreenNode?[]? newSlots = null;
+        for (var i = 0; i < green.SlotCount; i++)
+        {
+            var childGreen = green.GetSlot(i);
+            if (childGreen is null)
+                continue;
+
+            var newChild = childGreen.IsToken
+                ? RebuildToken(node, i, childGreen)
+                : node.GetNodeSlot(i) is { } childRed ? RebuildNode(childRed) : RebuildTokenList(node, i, childGreen);
```
meziantou: `RebuildNode` recurses once per level of the tree, which is the one walk in this layer that does. `GreenNode.WriteTo`, `GreenNode.IsEquivalentTo`, `SyntaxNodeRemover.RebuildEdge`, `SyntaxDiffer.Compare`, `SyntaxTreeDiagnostics.Walk` and `SyntaxNode.DescendantNodesAndTokensCore` all keep their own stack and say in their remarks that this is so a deeply nested document does not become a deep call stack.

The `CouldContainATarget` pruning above keeps the walk to the spine down to the target, so the recursion depth is the target's depth — still unbounded.

That depth is reachable, because the XML parser is iterative (`_elementStack`): `string.Concat(Enumerable.Repeat("<a>", 50_000))` parses without trouble, and only editing it fails. `root.ReplaceToken(...)`, `RemoveNode`, or `WithLeadingTrivia` (which goes through `ReplaceToken`) on a node near the bottom then recurses ~50,000 frames here and the process dies with a `StackOverflowException` the caller cannot catch. The JSON parser caps itself at `MaxDepth = 256`, but the XML, shell and regex parsers don't, so their trees can be deeper than this walk can rebuild.

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
meziantou: Fixed: `Splice` now takes the separator from the list when it holds one and otherwise from `parent.Green.CreateSeparator(slot)`, so the generated `CreateSeparator(int)` overrides are what answers here too -- `InsertNodesAfter` on a one-command pipeline gives `a|b`, not `a;b`. Test: `InsertNodesAfter_InAPipelineOfOneCommand_StillUsesAPipe`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/SyntaxNodeExtensions.cs
meziantou: Fixed: added `GreenNode.IsListSlot(int)` -- a general "does this slot hold a list", defaulting to `IsSeparatedListSlot` since a separated list is one, and generated from the `l`, `s` and `tl` slots (hand-written for JSON, XML and the test language). `SpliceIntoList` asks it when there is no index and refuses a single-node slot with the documented `ArgumentException`, so the member keeps its value instead of silently holding a `SyntaxListNode`. Tests: `InsertNodesAfter_WhereOnlyOneNodeFits_IsRejected` in the JSON and shared suites.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/Meziantou.Framework.Language/Syntax/SyntaxReplacer.cs
meziantou: Fixed: `RebuildNode` keeps its own stack of frames now, like the other walks here -- each frame holds the node, the slot the walk is at and the slots rebuilt so far, and a child's result comes back through `Accept`. A 10,000-deep document crashed the test host on both TFMs before the change and rebuilds fine after. Test: `ReplaceNode_InADeeplyNestedDocument_DoesNotOverflowTheStack`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #3111, opened 2026-09-11
Run tests on pull requests only and publish packages from a dedicated workflow

meziantou:
Pushes to main re-ran the whole test matrix (4 OSes × Debug/Release/invariant × shards) on code its pull request had already validated. This PR splits CI in two, by concern rather than by trigger.

## Changes

**`publish.yml`: new, the package pipeline**
- Runs on pull requests, on push to main, and manually.
- `create_nuget` (delta pack) → `validate_nuget` → `deploy`. Pull requests stop after validation: `deploy` only runs on a push to main, so a manual run from another branch can't publish either.
- Concurrency: a run on main gets a group of its own (`run_id`), because it packs only the packages changed since `github.event.before` and a cancelled or superseded run would never publish them. Pull request runs still cancel their predecessors.
- `workflow_dispatch` repacks every package and pushes with `--skip-duplicate`, to recover from a partial publish.
- feedz.io is gone: pull requests no longer push preview packages anywhere.

**`ci.yml`: pull request validation only**
- Renamed the workflow from `publish` to `ci`, removed the `push` trigger, and removed the `create_nuget`, `validate_nuget` and `deploy` jobs. It now covers eng scripts, generated files, analyzers, the test matrix and trimming.
- A manual run is still a full, non-delta run, with optional code coverage.
- `final_status_check` keeps its name, so the required branch-protection check still matches.

**`retry-ci.yml`**: watches both `ci` and `publish`.

## Before merging

- [ ] Add a nuget.org trusted-publishing policy for workflow file `publish.yml`. `NuGet/login` (OIDC) is tied to the workflow file name, so without it the first publish on main fails to log in. Remove the `ci.yml` policy after the first successful publish.
- [ ] Delete the now-unused `FEEDZ_APIKEY` repository secret.
- [ ] Consider making `create_nuget` and `validate_nuget` required checks: they now live in `publish.yml`, so `final_status_check` in `ci.yml` no longer covers them.

## Notes for reviewers

- The first `publish` run on main after merging repacks every package, because `publish.yml` is a full-rebuild trigger and the file is new. Pushes use `--skip-duplicate`, so that run is only slower.
- The publish path itself is first exercised by the merge, since `publish.yml` can't be dispatched before it exists on main. This PR's own runs exercise both workflows on the pull request side.
- Checked locally: both files parse as YAML, the deploy script passes the PowerShell parser, and `dotnet run ./eng/update-all.cs` reports no changes. actionlint doesn't support the `parallel:`/`case()` syntax these workflows use.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,12 +1,19 @@
+# Builds and validates the NuGet packages on pull requests, and publishes them on main.
 name: publish
 on:
   workflow_dispatch:
   push:
     branches:
       - 'main'
+  pull_request:
+    branches:
+      - '*'
 
-# No concurrency group: each run only packs the packages changed since the previous push, so a
-# cancelled or replaced run would never publish its packages.
+concurrency:
+  # A run on main gets a group of its own: it only packs the packages changed since the previous
+  # push, so a cancelled or replaced run would never publish them.
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_id || github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 env:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
@@ -25,7 +32,7 @@ jobs:
     outputs:
       has-project: ${{ steps.package-project.outputs.has-project }}
     env:
-      # Analyzers already ran on the pull request
+      # Analyzers run in the run_analyzers job of ci.yml
       RunAnalyzers: false
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
@@ -41,7 +48,7 @@ jobs:
           output-project: eng/packages.delta.proj
           base-commit: ${{ case(github.event_name == 'push', github.event.before, '') }}
           # A manual run packs every package, so it can recover from a partial publish
-          run-generate: ${{ case(github.event_name == 'push', 'true', 'false') }}
+          run-generate: ${{ case(github.event_name == 'pull_request' || github.event_name == 'push', 'true', 'false') }}
           full-rebuild-triggers: |
             .github/workflows/publish.yml
             eng/**/*.proj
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,193 @@
...
+          user: meziantou
+
+      - name: Publish NuGet packages
+        env:
+          NUGETAPIKEY: ${{steps.login.outputs.NUGET_API_KEY}}
+        run: |
+          Write-Host "Searching nupkg in folder: ${{ env.NuGetDirectory }}"
+          $files = Get-ChildItem "${{ env.NuGetDirectory }}/*" -Include *.nupkg
+          $apiKey = $env:NUGETAPIKEY
+          $source = "${{ env.NuGetSource }}"
```
meziantou: Use environment variables (`$env:`) to avoid value escaping issues

.github/workflows/publish.yml
```diff
@@ -0,0 +1,193 @@
...
+    outputs:
+      has-project: ${{ steps.package-project.outputs.has-project }}
+    env:
+      # Analyzers already ran on the pull request
+      RunAnalyzers: false
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          fetch-depth: 0
+      - run: git config --global protocol.file.allow always
```
meziantou: Not needed

.github/workflows/publish.yml
meziantou: Done in 0416b264f. Both PowerShell scripts in publish.yml now read `$env:NuGetDirectory` and `$env:NuGetSource` instead of interpolating the expressions, and the folder path is joined with `Join-Path` rather than string concatenation. The workflow-level `env:` entries are already exported to the shell, so no step-level `env:` was needed.

I left the `dotnet pack` line's `--output ${{ env.NuGetDirectory }}` alone: it is copied verbatim from ci.yml, and changing it only here would make the two files diverge. Happy to change it in both if you want.

.github/workflows/publish.yml
meziantou: Removed in 0a8ed0e82. Packing does not clone anything over the file protocol, so the step only mattered for the test jobs, which stay in ci.yml.

.github/workflows/publish.yml
meziantou: Now done in 0a3a14dd3: the pack step uses `--output $env:NuGetDirectory`, plus `$env:GITHUB_REF` and `$env:RUNNER_TEMP`. ci.yml no longer builds the packages at all, so there is no longer a copy of that line to keep in sync.

PR #3137, opened 2026-09-13
Add Meziantou.Framework.TaggedValues to detect mixed-up primitive values

meziantou:
Ids, keys, and units are often plain `Guid`, `int`, or `string` values, so nothing stops an order id from being compared with a project id, or passed where a project id is expected. Wrapper types prevent that, but each one needs converters for serializers, ORMs, and model binders.

This adds `Meziantou.Framework.TaggedValues`: tag the declarations, keep the primitive types, and let the bundled analyzer report when differently tagged values meet.

```c#
class Order
{
    [ValueTag("OrderId")] public Guid Id { get; set; }
    [ValueTag("ProjectId")] public Guid ProjectId { get; set; }
}

_ = order.Id == order.ProjectId;   // MFTV0001
Load(order.ProjectId);             // MFTV0002, Load([ValueTag("OrderId")] Guid orderId)

// ValueTag=OrderId
Guid id = ...;
var projectIdByOrderId = new Dictionary</* ValueTag=OrderId */ Guid, /* ValueTag=ProjectId */ Guid>();
```

Untagged values are never reported, so the package can be adopted one declaration at a time. An opt-in strict mode (`taggedvalues.strict = true`) also reports tagged values mixed with untagged values, except default values, `null`, constants, `Guid.Empty`, and new values such as `Guid.NewGuid()`.

## Tagging values

- `[ValueTag]` on fields, properties, parameters, and return values. Several tags declare a union; two values are compatible when they share a tag.
- Local variables use a comment: `/* ValueTag=OrderId */` next to the declaration, `// ValueTag=OrderId` on the line before it, or `/* ValueTag=OrderId */` in the type arguments of a `new` expression or of the declared type. Unannotated locals take the tag of their initializer.
- `[assembly: ValueTag(typeof(T), nameof(T.Member), "Tag")]` tags members of types you do not own, and is read from referenced assemblies too.
- Tags are inherited from overridden and implemented members, and from record primary constructor parameters.
- Naming conventions are opt-in (`taggedvalues.infer_tags_from_names = true`): `Id` on `Order` is `"OrderId"`, a parameter `Order id` is `"OrderId"`, and `orderId`, `_orderId`, and `s_orderId` are `"OrderId"`.

The tag of a collection, a `Nullable<T>`, a `Task<T>`, or a `Lazy<T>` describes its elements or its value, and `Key`/`Value` tag a dictionary. Tags flow through `foreach`, indexers, LINQ, lambda parameters (including `IQueryable` expression trees), anonymous type projections, and generic methods, by binding the type parameters of the member to the tags of its receiver and arguments. Built-in arithmetic keeps the tag (`orderId + 1`), while user-defined operators and conversions take the tag of their return value.

## Rules

| Id | Title |
| -- | -- |
| `MFTV0001` | Do not compare values with different tags |
| `MFTV0002` | Use a value with the tag the target expects |
| `MFTV0003` | Combine only values with the same tag |
| `MFTV0004` | Use the tags of the overridden or implemented member |
| `MFTV0005` | Fix or remove the invalid value tag annotation |
| `MFTV0006` | Add an explicit tag to disambiguate the conventional tag (conventions only) |
| `MFTV0007` | Remove the redundant value tag (conventions only) |
| `MFTV0008` | Tag the return value with the tag of the returned values (suggestion) |
| `MFTV0009` | Do not mix tagged values with untagged values (strict mode only) |

Messages name both declarations, their tags, and the line of the declaration to edit. Code fixes change the tag of the target of a mismatch or of an override (attribute or comment), add the suggested tag to a return value, and remove invalid or redundant annotations.

## Projects

- `src/Meziantou.Framework.TaggedValues`: the attribute (netstandard2.0, net10.0, net11.0), packing the analyzer and code fix into `analyzers/dotnet/cs`
- `src/Meziantou.Framework.TaggedValues.Analyzer` and `src/Meziantou.Framework.TaggedValues.CodeFix`
- `tests/Meziantou.Framework.TaggedValues.Analyzer.Tests`: 263 tests, including tags read from a referenced assembly's metadata, an internal copy of the attribute, exact message texts, and each code fix
- `readme.md`, `ref/`, root `README.md`, and `slnx` updated by `eng/update-all.cs`

## Validation

- `dotnet build -c Release /p:IsOfficialBuild=true /p:TreatWarningsAsErrors=true` for the four projects
- `dotnet test` on the analyzer tests: 263 passed
- `dotnet run ./eng/update-all.cs` leaves no changes
- Packed the package and built a separate console project against it: MFTV0001 and MFTV0002 are reported
- Ran the analyzer with naming conventions enabled on several large projects of this repository: no analyzer exception, well under a second per project


src/Meziantou.Framework.TaggedValues/Meziantou.Framework.TaggedValues.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <TargetFrameworks>$(LatestTargetFrameworks);netstandard2.0</TargetFrameworks>
+    <RootNamespace>Meziantou.Framework</RootNamespace>
+    <Version>1.0.0</Version>
+    <Description>Tag primitive values (ids, units, keys) with [ValueTag] and let the bundled Roslyn analyzer report when values with different tags are mixed</Description>
+    <PackageTags>analyzer, roslyn, strongly-typed-id, primitive-obsession, tagged-values, branded-types</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Remove="Meziantou.Polyfill" />
+  </ItemGroup>
```
meziantou: Remove

src/Meziantou.Framework.TaggedValues/Meziantou.Framework.TaggedValues.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Meziantou.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(LatestTargetFrameworks);netstandard2.0</TargetFrameworks>
+    <RootNamespace>Meziantou.Framework</RootNamespace>
```
meziantou: Remove and update namespaces accordingly

src/Meziantou.Framework.TaggedValues.Analyzer/TagInfo.cs
```diff
@@ -0,0 +1,197 @@
+using System.Collections.Immutable;
+using System.Text;
+
+namespace Meziantou.Framework.Analyzers.TaggedValues;
```
meziantou: The namespace doesn't match folder structure. Can you update all namespaces in this project.

src/Meziantou.Framework.TaggedValues.Analyzer/TagResolver.cs
```diff
@@ -0,0 +1,1185 @@
...
+            tags.Add(tag);
+        }
+
+        return tags.Aggregate(TagInfo.None, TagInfo.Union);
+    }
+
+    /// <summary>
+    /// Returns the first pair of combined values whose tags are incompatible.
+    /// </summary>
+    public bool TryFindIncompatibleValues(IEnumerable<IOperation> operations, out IOperation first, out TagInfo firstTags, out IOperation second, out TagInfo secondTags)
```
meziantou: Annotate the out parameters with NotNullWhen to reduce the number of null forgiving operator usages

src/Meziantou.Framework.TaggedValues.Analyzer/TagResolver.cs
```diff
@@ -0,0 +1,1185 @@
...
+    public static bool TryGetWrappedType(INamedTypeSymbol type, out ITypeSymbol wrappedType)
+    {
+        wrappedType = null!;
+        if (type.SpecialType is SpecialType.System_String)
+            return false;
+
+        if (type.TypeArguments.Length is 1)
+        {
+            var isWrapper = type.OriginalDefinition.SpecialType is SpecialType.System_Nullable_T or SpecialType.System_Collections_Generic_IEnumerable_T or SpecialType.System_Collections_Generic_IEnumerator_T ||
+                (type.Name is "Task" or "ValueTask" && IsInNamespace(type, "System", "Threading", "Tasks")) ||
```
meziantou: Get the symbols once in CompilationStart. Then, compare symbols

src/Meziantou.Framework.TaggedValues.Analyzer/TagResolver.cs
```diff
@@ -0,0 +1,1185 @@
...
+
+                default:
+                    return false;
+            }
+        }
+
+        return false;
+    }
+
+    private static bool IsInNamespace(ISymbol symbol, params string[] namespaceParts)
```
meziantou: Should not be needed as another comment suggests comparing symbols

src/Meziantou.Framework.TaggedValues.Analyzer/Meziantou.Framework.TaggedValues.Analyzer.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <NoWarn>$(NoWarn);RS1041</NoWarn>
+    <IncludeRoslynHelpers>true</IncludeRoslynHelpers>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.12.0" PrivateAssets="all" />
```
meziantou: Can you use version 5.0.0? (same for the codefix project)

tests/Meziantou.Framework.TaggedValues.Analyzer.Tests/NamingConventionTests.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    [Fact]
+    public async Task ConventionsInferTagsFromNames()
+    {
+        var source = Source
+            .Replace("[|Id == ProjectId|]", "{|MFTV0001:Id == ProjectId|}", StringComparison.Ordinal)
+            .Replace("[|_customerId == ProjectId|]", "{|MFTV0001:_customerId == ProjectId|}", StringComparison.Ordinal)
+            .Replace("[|ProjectId|]", "{|MFTV0002:ProjectId|}", StringComparison.Ordinal);
+
+        await VerifyAsync(source, inferTagsFromNames: true);
```
meziantou: Do not use the `Source` constant. Inline it, so there is no "Replace"

tests/Meziantou.Framework.TaggedValues.Analyzer.Tests/CollectionTagTests.cs
```diff
@@ -0,0 +1,222 @@
+namespace Meziantou.Framework.Tests;
+
+public sealed class CollectionTagTests : TaggedValuesAnalyzerTestBase
+{
+    private const string Declarations = """
```
meziantou: Do not use this pattern that have a large constant reused in multiple tests. Make each tests self-contained. Also, keep only the minimum for the tests to pass.

You can also split the tests into more smaller cases (more tests but smaller).

src/Meziantou.Framework.TaggedValues.Analyzer/TagResolver.cs
```diff
@@ -0,0 +1,1162 @@
...
+                if (name.Length is 0)
+                    return null;
+
+                return char.ToUpperInvariant(name[0]) + name.Substring(1);
+
+            case IPropertySymbol { IsIndexer: false } property:
+                return property.Name;
+
+            case IParameterSymbol parameter:
+                return parameter.Name.Length is 0 ? null : char.ToUpperInvariant(parameter.Name[0]) + parameter.Name.Substring(1);
```
meziantou: Casing is not important as tags use ignore case comparison, so non need to use ToUpperInvariant

